PR #17, opened 2020-08-08
KUBELET_DIR_PATH is required and can be changed as required

sstubbs:
Update the docs so people realise KUBELET_DIR_PATH is required and can be changed. Please see this issue minio/direct-csi#11

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This repository provides tools and scripts for building and testing the DIRECT C
 
 ```sh
 # Install the driver
-$> DIRECT_CSI_DRIVES=data{1...4} DIRECT_CSI_DRIVES_DIR=/mnt kubectl apply -k github.com/minio/direct-csi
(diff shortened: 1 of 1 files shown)


sstubbs:
I'm not entirely sure how to do that to be honest. I did try different variations before I did the PR to change it but it would either not read the environment variable or require the environment variable which is why I did it that way. Perhaps someone with a little more knowledge on this could fix that.

harshavardhana:
> I'm not entirely sure how to do that to be honest. I did try different variations before I did the PR to change it but it would either not read the environment variable or require the environment variable which is why I did it that way. Perhaps someone with a little more knowledge on this could fix that.

Yes in `resources/direct-csi.properties` the `KUBELET_DIR_PATH=` should default to `KUBELET_DIR_PATH=/var/lib/kubelet`

sstubbs:
I'll try it again. When I tried that it was overriding the env variable for some reason. I will create another pr for this if it works.

sstubbs:
Or actually I guess I will just update this branch as it's related to this and change the readme at the same time.

harshavardhana:
Fixed it properly with some doc cleanup https://github.com/minio/direct-csi/pull/18 feel free to close this PR

sstubbs:
ok great thanks

PR #44, opened 2020-11-13
Implement NodePublishVolume and Bugfixes

wlan0:
@Praveenrajmani @nitisht


pkg/apis/direct.csi.min.io/v1alpha1/types.go
```diff
@@ -61,7 +62,7 @@ type DirectCSIDrive struct {
 	// +optional
 	DriveStatus DriveStatus `json:"driveStatus,omitempty"`
 	// +optional
-	Topology *csi.Topology `json:"topology,omitempty"`
+	Topology map[string]string `json:"topology,omitempty"`
```
wlan0, line 65: Cannot use csi.Topology because DeepCopy is not implemented for it. Defaulting this to map[string]string

pkg/apis/direct.csi.min.io/v1alpha1/types.go
```diff
@@ -116,11 +117,15 @@ type DirectCSIVolume struct {
 	metav1.ObjectMeta `json:"metadata"`
 
 	// +optional
-	OwnerDrive *DirectCSIDrive `json:"ownerDrive,omitempty"`
+	OwnerDrive string `json:"ownerDrive,omitempty"`
```
wlan0, line 120: We should store a reference to the object (its name), rather than a copy of it. Otherwise we'll have multiple sources of truth

pkg/node/node.go
```diff
@@ -118,57 +116,131 @@ func (n *NodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetCa
...
 }
 
 func (n *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRequest) (*csi.NodeStageVolumeResponse, error) {
 	vID := req.GetVolumeId()
-	stagingTargetPath := req.GetStagingTargetPath()
-
 	if vID == "" {
 		return nil, status.Error(codes.InvalidArgument, "volume ID missing in request")
 	}
+	stagingTargetPath := req.GetStagingTargetPath()
```
wlan0, line 186: for idempotency. If it has already been staged. Then just respond with OK

pkg/node/node.go
```diff
@@ -118,57 +116,131 @@ func (n *NodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeGetCa
...
 		if cErr != nil {
 			return status.Error(codes.NotFound, cErr.Error())
 		}
-		csiDrive := dvol.OwnerDrive
+
+		if dvol.StagingPath == stagingTargetPath {
+			return nil
+		}
+
+		csiDrive, err := directCSIClient.DirectCSIDrives().Get(ctx, dvol.OwnerDrive, metav1.GetOptions{})
```
wlan0, line 203: pulling value from the reference.

pkg/node/node_controller.go
```diff
@@ -54,7 +55,11 @@ func (b *DirectCSIDriveListener) Delete(ctx context.Context, obj *v1alpha1.Direc
 }
 
 func startController(ctx context.Context) error {
-	ctrl, err := listener.NewDefaultDirectCSIController("node-controller", "leader-lock", 40)
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	ctrl, err := listener.NewDefaultDirectCSIController("node-controller", hostname, 40)
```
wlan0, line 62: this is to ensure that controllers from different nodes act as independent controllers, instead of a combined unit with a leader

pkg/node/utils.go
```diff
@@ -323,43 +322,3 @@ func WalkWithFollow(path string, callback func(path string, info os.FileInfo, er
 	}
 	return nil
 }
-
```
wlan0, line 326: moved it to its own file.

wlan0:
I've left comments for reviewers

Praveenrajmani (approved):

nitisht (approved):

PR #51, opened 2020-11-23
Replace uuids with sha256(nodeId-devicePath) for csi drive names

Praveenrajmani:
- This avoids creating new csidrives upon direct-csi (node) controller restarts
- Hash(NodeId-devicePath) will serve as a unique name for identifying csidrives

- Also, Clean up the go.mod by running `go mod tidy`


pkg/node/utils.go
```diff
@@ -174,7 +173,8 @@ func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_cs
 			}
 		}
 		v.OwnerNode = nodeID
-		driveName := uuid.NewUUID().String()
+		objName := strings.Join([]string{nodeID, v.Path}, "-")
+		driveName := fmt.Sprintf("%x", md5.Sum([]byte(objName)))
```
harshavardhana, line 177: instead of md5 use sha256 @Praveenrajmani to avoid any conflicts `sha256-simd` package can be used here. 

Also not sure why the variable name is `objName` isn't it `driveName` ?

Praveenrajmani: Will change @harshavardhana

harshavardhana:
Update the PR title @Praveenrajmani

nitisht (approved):
LGTM

wlan0 (approved):

PR #53, opened 2020-11-24
Add support for mountOptions in drive add command

nitisht:
Also update the drive list output table to show
servers in a separate column.


wlan0:
@nitisht looks good overall. You need to regenerate the CRDs using ./hack/update-codegen.sh

wlan0 (changes requested):
CRD spec needs to be regenerated using ./hack/update-codegen.sh

nitisht:
> CRD spec needs to be regenerated using ./hack/update-codegen.sh

Done

wlan0:
LGTM

config/crd/direct.csi.min.io_directcsidrives.yaml
```diff
@@ -15,85 +15,86 @@ spec:
...
+          path:
+            type: string
+          requestedFormat:
+            properties:
+              filesystem:
+                type: string
+              force:
+                type: boolean
+              mountoptions:
+                type: string
```
wlan0, line 72: this should be list

wlan0 (changes requested):

wlan0 (approved):

PR #57, opened 2020-11-29
fix(node): Fix symlink evaluation in FindDrives

haslersn:
The previous approach to evaluating symlinks (using `os.Readlink`) didn't take into account that symlinks can contain relative paths.

Note that I didn't test this commit, because I'm currently still having an issue where kubelet becomes unusable after deploying direct-csi. Furthermore, for me it seems like the whole `FindDrives` function and its helper functions need a major rework.

### DIFF
--- a/pkg/node/utils.go
+++ b/pkg/node/utils.go
@@ -20,17 +20,18 @@ import (
 	"bufio"
 	"context"
 	"fmt"
-	"github.com/golang/glog"
 	"io/ioutil"
-	"k8s.io/utils/mount"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"syscall"
 
+	"github.com/golang/glog"
 	direct_csi "github.com/minio/direct-csi/pkg/apis/direct.csi.min.io/v1alpha1"
 	simd "github.com/minio/sha256-simd"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/utils/mount"
 )
 
 func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_csi.DirectCSIDrive, error) {
@@ -69,26 +70,29 @@ func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_cs
 			return filepath.SkipDir
 		}
 
-		link, err := os.Readlink(path)
+		realpath, err := filepath.EvalSymlinks(path)
 		if err != nil {
-			link = path
+			// Following the symlink failed. Possibly the symlink dangled.
+			utilruntime.HandleError(err)
+			return nil
 		}
-		if _, ok := visited[link]; ok {
+
+		if _, ok := visited[realpath]; ok {
 			if info.IsDir() {
 				return filepath.SkipDir
 			}
 			return nil
 		}
-		visited[link] = struct{}{}
+		visited[realpath] = struct{}{}
 
 		// Find drive specific info
-		drive := getDriveForPath(drives, link)
+		drive := getDriveForPath(drives, path)
 		if drive == nil {
 			return nil
 		}
 		if strings.Compare(info.Name(), "wwid") == 0 {
 			if drive.SerialNumber == "" {
-				data, err := ioutil.ReadFile(link)
+				data, err := ioutil.ReadFile(path)
 				if err != nil {
 					return err
 				}
@@ -97,7 +101,7 @@ func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_cs
 		}
 		if strings.Compare(info.Name(), "model") == 0 {
 			if drive.ModelNumber == "" {
-				data, err := ioutil.ReadFile(link)
+				data, err := ioutil.ReadFile(path)
 				if err != nil {
 					return err
 				}
@@ -109,7 +113,7 @@ func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_cs
 			if drive.Name == drive.RootPartition {
 				return nil
 			}
-			data, err := ioutil.ReadFile(link)
+			data, err := ioutil.ReadFile(path)
 			if err != nil {
 				return err
 			}
@@ -120,7 +124,7 @@ func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_cs
 			drive.PartitionNum = partNum
 		}
 		if strings.Compare(info.Name(), "size") == 0 {
-			data, err := ioutil.ReadFile(link)
+			data, err := ioutil.ReadFile(path)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


pkg/node/utils.go
```diff
@@ -69,26 +70,29 @@ func FindDrives(ctx context.Context, nodeID string, procfs string) ([]*direct_cs
 			return filepath.SkipDir
 		}
 
-		link, err := os.Readlink(path)
+		realpath, err := filepath.EvalSymlinks(path)
```
harshavardhana, line 73: EvalSymlinks is the correct usage @haslersn thanks for the fix

wlan0: It doesnt add any value because absolute paths are not necessary here. How is this the correct fix?

haslersn: It is necessary to use the realpath in `visited` because otherwise there might be an endless loop in case of cyclic symlinks (or rather it fails at some depth where there are too many symlinks).

wlan0:
There is no value in memoizing the absolute paths. What’s the problem with findDrives? 

If you’re willing to help, we could use your skills for detecting the filesystem of a device by reading its super block. Would you be able to help with that? 

Additionally, what is the problem with your kubelet when deploying direct-csi?

haslersn:
I think memorizing absolute paths is necessary in order to prevent otherwise possible endless recursions.

Yes, I think I can look into FS detection. First I do however want to look at the drive detection (i.e. `FindDrives`) because after reading the code I think there might be some edge cases where it doesn't work out right.

The problem with my kubelet is that it doesn't terminate or create pods anymore. Pods that are attempted to be terminated or created are stuck in `Terminating` or `ContainerCreating` state, respectively. All of my 3 nodes are affected by this problem and it only ever happens after I deploy direct-csi. It can be fixed by deleting direct-csi (The pods can't be terminated, of course, as I stated.) and then rebooting all 3 nodes.

haslersn:
> What’s the problem with findDrives?

One thing I can't wrap my head around is: Why does `FindDrives` first walk arbitrarily deep into the `/sys/block/` subtree and then call `getPartition` which, judging from the code, only supports block devices listed in the top-level (like `/sys/block/sda`) or second-level (like `/sys/block/sda/sda1`). If all of the relevant block devices are at those two levels, then I think the code can be cleaned up greatly by simply listing `/sys/block/` in order to find devices and then listing `/sys/block/devicename` for each device in order to find partitions.

Also, I think that the "primary key" of a `DirectCSIDrive` resource should be one of its IDs with which it is listed in `/dev/disk/by-id`. What do you think?

pkg/node/utils.go
harshavardhana: The problem is cyclical links are possible with symlinks - ReadLink is not designed for that https://golang.org/pkg/path/filepath/#EvalSymlinks is the correct from an expectation point of view.

wlan0: I'm well aware of cyclical links. In this case, since we are not recursing unboundedly, it is impossible. Now, if it was possible that there were cyclical links. EvalSymlinks would still not help. Because the in visited[path], the path, which would be a result of EvalSymlinks is still a relative path. 

I'm not sure what prompted this PR, since it does not fix a problem. If the kubelet is thrashing because of recursive calls from a completely different process, which would run out of memory if this infinite recursion happened, and not affect the kubelet. There is no way direct-csi would affect kubelet directly.

Just to be thorough. We have been testing this on mutliple environments as a part of our testing to ensure that it works with multiple versions of kubernetes. This issue has never occurred.

wlan0:
> I think memorizing absolute paths is necessary in order to prevent otherwise possible endless recursions.
I'm talking about memoization. There are no endless recursions. Endless recursion would quickly run out of stack memory and get OOM killed. The probelm you're experiencing has a completely different profile, which as of now, I cannot tell how direct-csi would be responsible. 

> The problem with my kubelet is that it doesn't terminate or create pods anymore. Pods that are attempted to be terminated or created are stuck in Terminating or ContainerCreating state, respectively. All of my 3 nodes are affected by this problem and it only ever happens after I deploy direct-csi. It can be fixed by deleting direct-csi (The pods can't be terminated, of course, as I stated.) and then rebooting all 3 nodes.

This could be because of a different issue. I'm happy to do a combined debugging session. I would like to learn if there are some weird states being produced by direct-csi. Though direct-csi should not affect another completely different process i.e. kubelet.

wlan0:
> One thing I can't wrap my head around is: Why does FindDrives first walk arbitrarily deep into the /sys/block/ subtree and then call getPartition which, judging from the code, only supports block devices listed in the top-level (like /sys/block/sda) or second-level (like /sys/block/sda/sda1). If all of the relevant block devices are at those two levels, then I think the code can be cleaned up greatly by simply listing /sys/block/ in order to find devices and then listing /sys/block/devicename for each device in order to find partitions.

If you list the top level, then you cannot find out partitions. You'll have to know the exact filetree for finding relevant information across multiple implementations of linux kernels. It is important to use automatic detection of drives rather than expecting a specific filetree because not all sysfs are created equal. This method however covers all the recent versions, if not all the versions of linux kernel.

> Also, I think that the "primary key" of a DirectCSIDrive resource should be one of its IDs with which it is listed in /dev/disk/by-id.  This way it doesn't depend on the order in which the init system recognizes drives. What do you think? And then, why not just  
list /dev/disk/by-id/ in order to find drives?

The primary key need not be the disk id. In case of drives being replaced, we do not want to consider a new drive mount at the same path as a new drive. We want to rather consider them as the same drive as before. The by-id symlinks are just `wwid`, which we already capture. On startup, we do plan on adding a check if drives are detected out of order, and update the CRDs according.

haslersn:
> There are no endless recursions.

Conceptually it is an endless recursion, but in reality my direct-csi crashes because some file system operation on the `path` fails as soon as the path involves too many symlinks. In my case:

- `/sys/block/nvme0n1/nvme0n1p1/holders/dm-0` links to `/sys/devices/virtual/block/dm-0`
- `/sys/devices/virtual/block/dm-0/slaves/nvme0n1p1` links back to nvme0n1p1

So the walk recurses until something fails.

(All of this has nothing to do with my kubelet problems. They also happen with this fix applied.)

(All of this has nothing

wlan0:
I see, I would like to learn more about your environment. Could we get on a zoom call?

pkg/node/utils.go
harshavardhana: From what I understood from this PR - this was a defensive change @wlan0 - not sure if there is a valid error that this PR fixes perhaps @haslern should fill us in. 

It is important to note that ReadLink should never be used in a naked manner, generally it is known to cause cyclical loops - EvalSymlinks simply traverses everything to give you a path that doesn't result in ELOOP - i.e ELOOP is returned if indeed that is going to happen eventually - ReadLink, on the other hand, it won't return such an error until you reach a weird state recursively going deeper.

So from what I saw it is a defensive code to avoid naked ReadLink usage.

harshavardhana:
> * `/sys/block/nvme0n1/nvme0n1p1/holders/dm-0` links to `/sys/devices/virtual/block/dm-0`
> * `/sys/devices/virtual/block/dm-0/slaves/nvme0n1p1` links back to nvme0n1p1

IMO dm-0's should be ignored by direct-csi and use directly the drives that are not part of any RAID/LVM setup.

wlan0:
@harshavardhana We are discussing on a call right now

wlan0:
@harshavardhana you're right. There are a few other issues with this PR that I've discussed with sebastian. He's going to make new PRs to fix them

PR #58, opened 2020-11-29
feat(node): Rework FindDrives

haslersn:
Now the DirectCSIDrives' names are of the form

```go
fmt.Sprintf("%s-%s", nodeID, serialNumber)
```

where `serialNumber` is an ID of the drive as in `/dev/disk/by-id`. We prefer shorter IDs (with lower lexicographic order being the tie-breaker). If the ID is not a valid DNS-1123 subdomain, then it is (in the `DirectCSIDrive` name, only) converted to a valid one by replacing illegal characters.

Signed-off-by: Sebastian Hasler <sebastian.hasler@sec.uni-stuttgart.de>

Closes #57

PR #62, opened 2020-11-30
Revert "feat(node): Rework FindDrives (#58)" and bugfixes

Praveenrajmani:
This reverts commit 511459038721d521a425e4f8d547651889271598.

listing disks by id is inconsistent among different versions of sysfs

Note: Also fixes the drive's topology by adding region

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -43,7 +43,7 @@ func NewNodeServer(ctx context.Context, identity, nodeID, rack, zone, region str
 	topologies[topology.TopologyDriverIdentity] = identity
 	topologies[topology.TopologyDriverRack] = rack
 	topologies[topology.TopologyDriverZone] = zone
-	topologies[topology.TopologyDriverZone] = zone
+	topologies[topology.TopologyDriverRegion] = region
 	topologies[topology.TopologyDriverNode] = nodeID
 	driveTopology := topologies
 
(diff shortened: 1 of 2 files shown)


haslersn:
Identifying disks by ID (e.g. `/dev/disk/by-id/nvme-eui.00253859915001a2`) is more consistent than identifying them by device path (e.g. `/dev/disk/nvme0n1`). So are you proposing to go from a pretty consistent (and maybe not perfect) method back to a pretty inconsistent method?

As far as I understand, your concern is only about the `DirectCSIDrives.ObjectMeta.Name`, right? Then there's no need to revert the full commit. Instead we could just change the name into something that's we can agree is more consistent. But what would that be?

haslersn:
One possibility is to require that drives have a WWN and use that WWN. If it's ok to only support disks that have a WWN (and their partitions), then that seems to be the way to go.

Praveenrajmani:
Yeah @haslersn 

Based on a discussion with @wlan0 . The current drive-discovery mechanism is dependent on the underlying sysfs which may vary based on the versions.. For eg, We may get errors like `Error: open /dev/disk/by-id: no such file or directory` as id based listing is not available in few versions.. We faced this issue while testing..

haslersn:
The old code had a problem with recursion depth where it crashed for me. So by reverting, you trade one bug by another.

If there indeed exist systems where `/dev/disk/by-id` does not exist, then we can't use that. We must find a different way to uniquely and consistently identify drives.

wlan0:
Yeah @haslersn @Praveenrajmani, I'm going to explore ioctl today, and get back to you with more info

PR #63, opened 2020-11-30
Add finalizers to DirectCSI drives

Praveenrajmani:
- Adds a finalizer <volume-id> to the csi drive when a volume is created

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -143,10 +143,11 @@ func (c *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		copiedDrive.Status.FreeCapacity = copiedDrive.Status.FreeCapacity - req.GetCapacityRange().GetRequiredBytes()
 		copiedDrive.Status.AllocatedCapacity = copiedDrive.Status.AllocatedCapacity + req.GetCapacityRange().GetRequiredBytes()
 		copiedDrive.Status.DriveStatus = direct_csi.Online
+		copiedDrive.ObjectMeta.Finalizers = utils.AddFinalizer(copiedDrive.ObjectMeta.Finalizers, vol.ObjectMeta.Name)
 		if _, err := directCSIClient.DirectCSIDrives().Update(ctx, copiedDrive, metav1.UpdateOptions{}); err != nil {
 			return err
 		}
-		glog.Infof("Updated DirectCSI DirectCSIDrive - %s", copiedDrive.Name)
+		glog.Infof("Updated DirectCSIDrive - %s", copiedDrive.Name)
 
 		return nil
 	})
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/glog"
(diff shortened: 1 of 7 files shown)


haslersn:
Is this about volumes being automatically deleted when you delete a drive? If yes, then couldn't you instead add an `OwnerReference` to the volume that points to the drive. Combined with `blockOwnerDeletion=true` Kubernetes will ensure that the drive resource is still available while the volumes are being deleted.

Praveenrajmani:
> Is this about volumes being automatically deleted when you delete a drive? 
Yes, The ownerReference might fix the automatic deletion problem.

The finalizers here is for the purging use case which is yet to come.. We should not not allow deletes on a drive which is been tied (in-use) to a direct csi volume.

haslersn:
Ah, I see. Can a finalizer prohibit deletion? Or is this something for a validating admission web hook?

Praveenrajmani:
We are basically trying to do this - https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers

haslersn:
I know what a finalizer is. Maybe I misunderstood, but it seems that you are trying to disallow deletion:

> We should not not allow deletes on a drive which is been tied (in-use) to a direct csi volume.

However, a finalizer is not meant for disallowing the deletion. It is meant for running some logic before the object is deleted.

So if you want to delete volumes when the drive is requested to be deleted, then that's a use case for a finalizer. But if you want to disallow deleting drives that still have volumes, then that is **not** a use case for a finalizer.

wlan0:
> However, a finalizer is not meant for disallowing the deletion. It is meant for running some logic before the object is deleted.

That is not true. OwnerReference forces a parent child relationship between objects, and can easily lead to orphaned children. It is meant for resources that lead to the creation of another resource. For instance, creating a `service` resource, leads to the creation of `endpoints` resource. In our case, it doesn't make sense for the use case that Praveen is making a PR. 

We are using finalizer for two reasons:
1. To wait for volumes of a particular drive to go to Offline status
2. To delete volumes when the drive is deleted (assuming Purge is set, and force is set)

haslersn:
I think you misunderstood. Your two reasons are about deferring the deletion of a drive and not about disallowing its deletion. Eventually the drive is deleted.

If you want to disallow deletion (i.e. the API user should get an error if trying to delete a drive), then finalizer is the wrong tool.

haslersn:
> OwnerReference forces a parent child relationship between objects, and can easily lead to orphaned children

Not if you enable cascading deletion.

wlan0:
Cascading deletion is a client side operation and requires extra handling if the client misses this step.

Praveenrajmani:
> To wait for volumes of a particular drive to go to Offline status
This is exactly the intention of this PR (To reword the previous comment)

haslersn:
> Cascading deletion is a client side operation

The [Kubernetes doc says that cascading deletion is performed by the Kubernetes garbage collector](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#background-cascading-deletion) which apparently [is part of the kubelet](https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/), so how can that be a client side operation?

pkg/utils/clients.go
```diff
@@ -69,3 +69,26 @@ func GetDirectCSIClient() directv1alpha1.DirectV1alpha1Interface {
 func GetCRDClient() apiextensions.CustomResourceDefinitionInterface {
 	return crdClient
 }
+
+func AddFinalizer(finalizers []string, finalizer string) []string {
```
wlan0, line 73: https://github.com/kubernetes/apimachinery/blob/681a08151eac875afc5286670195105118d3485d/pkg/api/meta/meta.go#L593

wlan0:
I can see that you're only adding finalizers. Where are they being removed. Is this PR only for a part of the lifecycle?

Praveenrajmani:
> this PR only for a part of the lifecycle

Yes. Will be doing that in the next PR - https://github.com/minio/direct-csi/pull/64

pkg/controller/controller.go
```diff
@@ -143,10 +143,11 @@ func (c *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolu
 		copiedDrive.Status.FreeCapacity = copiedDrive.Status.FreeCapacity - req.GetCapacityRange().GetRequiredBytes()
 		copiedDrive.Status.AllocatedCapacity = copiedDrive.Status.AllocatedCapacity + req.GetCapacityRange().GetRequiredBytes()
 		copiedDrive.Status.DriveStatus = direct_csi.Online
+		copiedDrive.ObjectMeta.Finalizers = utils.AddFinalizer(copiedDrive.ObjectMeta.Finalizers, vol.ObjectMeta.Name)
```
haslersn, line 146: I think here you don't follow the [naming convention for finalizers](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers):

> Identifiers of custom finalizers consist of a domain name, a forward slash and the name of the finalizer.

haslersn (changes requested):

pkg/controller/controller.go
haslersn:
```suggestion
		copiedDrive.ObjectMeta.Finalizers = utils.AddFinalizer(copiedDrive.ObjectMeta.Finalizers, fmt.Sprintf("%s/%s, v1alpha1.SchemeGroupVersion.Group, vol.ObjectMeta.Name))
```

wlan0: This is a good catch. Please look into this: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/validation/objectmeta.go#L103

wlan0 (approved):

PR #66, opened 2020-12-01
fix(scripts): Take bash from PATH

haslersn:
Not all distributions have `bash` under `/bin/bash` (such as NixOS), but usually they do at least have `env` under `/usr/bin/env` because I think `/usr/bin/env` is needed for POSIX compliance.


wlan0:
You could just change this to sh

wlan0 (approved):

PR #70, opened 2020-12-04
Add more robust drive discovery

wlan0:
@Praveenrajmani


wlan0:
@haslersn

pkg/dev/dev.go
```diff
@@ -0,0 +1,128 @@
...
+		if strings.HasPrefix(info.Name(), "loop") {
+			return filepath.SkipDir
+		}
+		if info.Name() != "uevent" {
+			return nil
+		}
+		drive, err := parseUevent(path)
+		if err != nil {
+			glog.V(10).Info(err)
+			return nil
```
Praveenrajmani, line 54: Good that we dont error out on parse failure :+1

Praveenrajmani (approved):

pkg/dev/dev.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+)
+
+type BlockDevice struct {
```
haslersn, line 30: We could also need to store the devpath in order to query more information later. The devpath is the path to the device directory relative to `/sys`, without any symlink elements and with leading `/`. It is used by the kernel as a unique key to the device. Example devpath:

```
/devices/pci0000:00/0000:00:1d.4/0000:3d:00.0/nvme/nvme0/nvme0n1
```

pkg/dev/dev.go
haslersn: Does this happen under normal operation or is it exceptional? If it's exceptional, then maybe log an `Error` instead of `Info`.

haslersn:
Whats your plan on reading attributes such as the following?

Attributes we probably want:

- Block device size
- File system type
- File system UUID
- File system capacity (This might differ a bit from the block device size.)
- File system free capacity

Drive attributes that are useful as information for the user, but are not required and not always present:

- Vendor
- Model
- Serial number
- WWN

Partition attributes that are useful as information for the user:

- Devname of the drive

pkg/dev/dev.go
wlan0: It’s regular

pkg/dev/dev.go
wlan0: There is no value in it. This is the physical port where the drive is attached. If the admin changes the slots, then this path will change but it won’t matter if we use the file system uuid.

wlan0:
The concerns you’ve raised are separate from discounting the drives itself. This PR only addresses drive discovery

pkg/dev/dev.go
haslersn: Of course the value is not constant for the physical drive, but that was not my intention. My intention is that you might need the devpath in order to query the kernel API for the attributes I mentioned in my other comment.

pkg/dev/dev.go
```diff
@@ -0,0 +1,128 @@
...
+	Devtype string
+}
+
+func FindDrives(ctx context.Context) ([]BlockDevice, error) {
+	const head = "/sys/devices"
+	drives := []BlockDevice{}
+
+	if err := filepath.Walk(head, func(path string, info os.FileInfo, err error) error {
+		if ctx.Err() != nil {
+			return err
```
haslersn, line 43: Shouldn't this be

```suggestion
			return ctx.Err()
```

PR #72, opened 2020-12-04
Build kubectl-directcsi as a single binary along with direct-csi

Praveenrajmani:
- Use `go build -tags "plugin" -o kubectl-directcsi` to build the plugin
- Use `go build` to build direct-csi


nitisht (changes requested):
LGTM overall, but shouldn't you add `// +build plugin` to plugin specific files for the `-tags "plugin"` to work?

Praveenrajmani:
> LGTM overall, but shouldn't you add `// +build plugin` to plugin specific files for the `-tags "plugin"` to work?

Yeah, Makes sense.

wlan0 (approved):

PR #79, opened 2020-12-10
probe mountinfo for each blockdevice

wlan0:
The format of mountinfo is guaranteed to be stable across kernel versions. See https://man7.org/linux/man-pages/man5/proc.5.html  (search for /proc/[pid]/mountinfo)

### DIFF
--- a/pkg/dev/mount.go
+++ b/pkg/dev/mount.go
@@ -1,0 +1,116 @@
+// This file is part of MinIO Direct CSI
+// Copyright (c) 2020 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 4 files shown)


pkg/dev/mount.go
```diff
@@ -0,0 +1,116 @@
...
+				return nil, fmt.Errorf("invalid format of %s 5", mountinfoFile)
+			}
+
+			fsType := FSType(secondParts[0])
+			mountSource := secondParts[1]
+			if !strings.HasSuffix(mountSource, fmt.Sprintf("%d", partitionNum)) {
+				continue
+			}
+			superblockOptions := strings.Split(secondParts[2], ",")
+			fmt.Println("found mount", line)
```
Praveenrajmani, line 100: We can remove this print

Praveenrajmani (approved):
LGTM otherwise

pkg/dev/mount.go
Praveenrajmani: Will remove this in the next PR

PR #84, opened 2020-12-10
Find device and partition paths while discovering drives

Praveenrajmani:
- If the device file is not found, make device file using `mknod`
- Extract partition paths from dentries for `/dev`


pkg/dev/partitions.go
```diff
@@ -95,6 +98,13 @@ func (b *BlockDevice) FindPartitions() ([]Partition, error) {
 		if partType == "" {
 			partType = partTypeUUID
 		}
+
+		partNum := int(i + 1)
+		partitionPath := b.Path + "p" + strconv.Itoa(partNum)
```
wlan0, line 103: remember we talked about not having the prefix 'p'

wlan0: it might be better to even do `-part-{1...n}`

wlan0 (approved):

PR #98, opened 2020-12-15
Implement NodeGetVolumeStats RPC and bugfixes

Praveenrajmani:
- Add xfs_quota support to set the hard limits for a volume
- Fix the formatting to cover only xfs


Praveenrajmani:
Tested the feature. The PR is ready for the review.

Dockerfile
```diff
@@ -13,6 +13,7 @@ FROM alpine:latest
 
 WORKDIR /
 RUN apk add xfsprogs
+RUN apk add xfsprogs-extra
```
wlan0, line 16: This should not be needed?

Dockerfile
```diff
@@ -13,6 +13,7 @@ FROM alpine:latest
...
 RUN apk add e2fsprogs
```
wlan0, line 17: This can be removed

Dockerfile
```diff
@@ -13,6 +13,7 @@ FROM alpine:latest
...
 RUN apk add dosfstools
```
wlan0, line 18: This can be removed

Dockerfile
wlan0: Nvm, this is this for the quota right?

pkg/node/publish_unpublish.go
```diff
@@ -48,7 +48,7 @@ func PublishVolume(ctx context.Context, stagingPath, containerPath string, readO
 	}
 
 	if shouldBindMount {
-		opts := []string{"bind"}
+		opts := []string{"bind", "prjquota"}
```
wlan0, line 51: What happens if there is no storage requirement specified on the requested volume?

wlan0 (commented):
Minor comments.

Dockerfile
Praveenrajmani: yes

pkg/node/publish_unpublish.go
Praveenrajmani: As we run xfs_quota on these mounts, We will add this option whenever we do a bind mount.

wlan0 (approved):

PR #99, opened 2020-12-16
Unmount the container and staging paths of the volumes during deletion

### LINKED ISSUE #90: Unmount volume on delete

Praveenrajmani:
Fixes #90

### DIFF
--- a/pkg/utils/clients.go
+++ b/pkg/utils/clients.go
@@ -17,6 +17,8 @@
 package utils
 
 import (
+	"path/filepath"
+
 	directv1alpha1 "github.com/minio/direct-csi/pkg/clientset/typed/direct.csi.min.io/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/utils/clients.go
```diff
@@ -116,3 +119,25 @@ func CheckVolumeStatusCondition(statusConditions []metav1.Condition, condType st
 	return false
 
 }
+
+// UnmountIfMounted - Idempotent function to unmount a target
+func UnmountIfMounted(mountPoint string) error {
```
wlan0, line 124: I'm not sure if this UnmountAll behavior should be the default. Regardless, lets move forward with this, and address any issues as they arise

Praveenrajmani: We use `UnmountAllRefs` only while unmounting & formatting (add-drives)

wlan0 (approved):

PR #117, opened 2020-12-22
Refresh DirectCSIVolume stats on volume updates

Praveenrajmani:
Also includes DirectCSIVolume type refactorings


pkg/apis/direct.csi.min.io/v1alpha1/types.go
```diff
@@ -134,6 +134,13 @@ type DirectCSIVolume struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
 
+	Status DirectCSIVolumeStatus `json:"status,omitempty"`
+	// +listType=atomic
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
```
wlan0, line 140: Conditions should also go under status

Dockerfile
```diff
@@ -1,11 +1,17 @@
+FROM golang:1.14 AS build
```
wlan0, line 1: Do we need this change?

pkg/controller/volume_controller.go
```diff
@@ -54,14 +56,36 @@ func (b *DirectCSIVolumeListener) Add(ctx context.Context, obj *v1alpha1.DirectC
 func (b *DirectCSIVolumeListener) Update(ctx context.Context, old, new *v1alpha1.DirectCSIVolume) error {
 	glog.V(1).Infof("Update called for DirectCSIVolume %s", new.ObjectMeta.Name)
 
+	directCSIClient := b.directcsiClient.DirectV1alpha1()
+	if utils.CheckVolumeStatusCondition(new.Conditions, "published", metav1.ConditionTrue) {
+		sc := utils.GetVolumeStatusCondition(new.Conditions, "volumestats")
+		duration := time.Since(sc.LastTransitionTime.Time)
+		if duration.Minutes() >= float64(5) {
```
wlan0, line 63: if duration > 5 * time.Minute {

resources/workloads.yaml
```diff
@@ -97,7 +97,7 @@ spec:
         - name: direct-csi
           image: $(REPOSITORY_ORG)/$(REPOSITORY_IMAGE):$(VERSION)
           args:
-            - "--identity=direct.csi.min.io"
```
wlan0, line 100: Is this because plugin and kustomize yaml have diverged?

Praveenrajmani: yes

wlan0 (approved):

PR #125, opened 2020-12-28
[Need to test] Introduce a namespace-lock to mount/umount the staged/published paths

Praveenrajmani:


pkg/node/nslock.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+// Lock the namespace resource.
+func (n *nsLockMap) lock(ctx context.Context, resource string) bool {
+	n.lockMapMutex.Lock()
+	defer n.lockMapMutex.Unlock()
+
+	// Check if the resource is locked
+	isLocked, found := n.lockMap[resource]
```
wlan0, line 63: {
  if _, ok := n.lockMap[resource]; ok {
       return false
  }
  n.lockMap[resource] = true
  return true
}

Praveenrajmani:
Closing this PR as we decided to have a simpler locking scheme.

PR #126, opened 2020-12-29
Introduce a path-lock to mount/umount the staged/published paths

Praveenrajmani:


cmd/direct-csi/cmd.go
```diff
@@ -67,6 +67,7 @@ For more information, use '%s man [sched | examples | ...]'
 
 func init() {
 	if driverCmd.Version == "" {
+		Version = "dev"
```
Praveenrajmani, line 70: Should fix this

wlan0 (approved):

PR #127, opened 2021-01-04
Move periodic volume stat updates to node controller

Praveenrajmani:
The volume controller instances will not be running on all nodes and running xfs_quota reports on volume controller is not usefull in collecting stats. 

volume stat updates are specific to node controller so this PR spins a thread on all the nodes to refresh the stats with a ticker of 5mins.

### DIFF
--- a/pkg/node/drive_controller.go
+++ b/pkg/node/drive_controller.go
@@ -198,39 +198,35 @@ func startDriveController(ctx context.Context, nodeID string) error {
 	return ctrl.Run(ctx)
 }
 
-func refreshVolumeStats(ctx context.Context, nodeID string) {
-	retryTicker := time.NewTicker(5 * time.Minute)
-	defer retryTicker.Stop()
+func refreshVolumeStats(ctx context.Context, nodeID string, refreshTimeInterval time.Duration) {
+	refreshTicker := time.NewTicker(refreshTimeInterval)
+	defer refreshTicker.Stop()
 
 	directCSIClient := utils.GetDirectCSIClient()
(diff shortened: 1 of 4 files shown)


pkg/node/drive_controller.go
```diff
@@ -192,3 +197,74 @@ func startDriveController(ctx context.Context, nodeID string) error {
...
+	retryTicker := time.NewTicker(5 * time.Minute)
+	defer retryTicker.Stop()
+
+	directCSIClient := utils.GetDirectCSIClient()
+	refreshStatsFunc := func(dvol direct_csi.DirectCSIVolume) error {
+		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if utils.CheckVolumeStatusCondition(dvol.Status.Conditions, "published", metav1.ConditionTrue) {
+				sc := utils.GetVolumeStatusCondition(dvol.Status.Conditions, "volumestats")
+				duration := time.Since(sc.LastTransitionTime.Time)
+				if duration > 5*time.Minute {
```
Praveenrajmani, line 211: This condition is not required anymore, Will remove this

pkg/node/drive_controller.go
```diff
@@ -192,3 +197,74 @@ func startDriveController(ctx context.Context, nodeID string) error {
 	ctrl.AddDirectCSIDriveListener(&DirectCSIDriveListener{nodeID: nodeID})
 	return ctrl.Run(ctx)
 }
+
+func refreshVolumeStats(ctx context.Context, nodeID string) {
+	retryTicker := time.NewTicker(5 * time.Minute)
```
Praveenrajmani, line 202: Will introduce a constant for `retryInterval`

PR #128, opened 2021-01-05
Setup Validation admission controllers for DirectCSI drives and volumes

Praveenrajmani:
- Add the following rules in drive validation

```
 - Check if the fstype in the requestedFormat == "xfs"
 - Check if directCSIOwned is not set to True or requestedFormat is set for root partitions (unavailable drives)
 - Check if requestedFormat is not set for a drive in-use
 - Check if force option is set if the drive has an existing filesystem or mountpoint
```

- Also, Register and unregister drive validation controller and rules during install/un-install direct-csi commands
- Create a controller service for the central-controller deployment for the validation webhook server

### DIFF
--- a/cmd/kubectl-direct_csi/uninstall.go
+++ b/cmd/kubectl-direct_csi/uninstall.go
@@ -145,6 +145,20 @@ func uninstall(ctx context.Context, args []string) error {
 	}
 	glog.Infof("'%s' daemonset deleted", utils.Bold(identity))
 
+	if err := installer.DeleteDriveValidationRules(ctx, identity); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+	glog.Infof("'%s' drive validation rules removed", utils.Bold(identity))
+
+	if err := installer.DeleteSecrets(ctx, identity); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+	glog.Infof("'%s' secrets deleted", utils.Bold(identity))
+
 	if err := installer.DeleteDeployment(ctx, identity); err != nil {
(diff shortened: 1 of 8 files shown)


cmd/kubectl-direct_csi/uninstall.go
```diff
@@ -145,6 +145,20 @@ func uninstall(ctx context.Context, args []string) error {
...
+		}
+	}
+	glog.Infof("'%s' drive validation rules removed", utils.Bold(identity))
+
+	if err := installer.DeleteSecrets(ctx, identity); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+	glog.Infof("'%s' secrets deleted", utils.Bold(identity))
```
wlan0, line 160: Please replace this line with the log ling from L#153 here

pkg/utils/installer/install.go
```diff
@@ -25,11 +25,14 @@ import (
...
 	"github.com/minio/direct-csi/pkg/utils"
 
+	admissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kErr "k8s.io/apimachinery/pkg/api/errors"
```
wlan0, line 35: This is a nit - can we make this `kerr` instead of `kErr`. I'll merge this PR, but as a part of the next PR, we should take care of this

wlan0 (approved):

PR #137, opened 2021-01-19
unmount all mounts for a particular drive and doc updates

wlan0:


docs/architecture.md
```diff
@@ -0,0 +1,39 @@
...
+
+The 4 components run as two different pods. 
+
+| Name                          | Components                                           | Description                        |
+|-------------------------------|------------------------------------------------------|------------------------------------|
+| DirectCSI Node Driver         | CSI Driver, Driver Controller, Volume Controller     | runs on every node as a DaemonSet  |
+| DirectCSI Central Controller  | CSI Controller                                       | runs as a deployment               |
+
+
+### Scalability
```
Praveenrajmani, line 27: Can we say that the default replica count for `DirectCSI Central Controller` is 3?

docs/index.md
```diff
@@ -9,13 +9,16 @@ Documentation
  - [Installation](./installation.md)
  - [CLI reference](./cli.md)
 
+### Advanced
+ - [Architecture](./architecture.md)
+ - [Security](./security.md)
+ - [CSI Specification](./specification.md)
+
```
Praveenrajmani, line 16: Can we also mention the minimum version requirements? Like, Kubernetes version, sidecar container versions [csi-node-driver-registrar, csi-provisioner, csi-driver, etc]

wlan0: Good point

wlan0: I've added it to installation docs

PR #138, opened 2021-01-19
Add pod labels to the DirectCSIVolume while publishing volume and BugFixes

Praveenrajmani:
Attach the pod labels prefixed with `direct.csi.min.io` such as pod names and namespaces to the DirectCSIVolume, Which can be used to filter the volume listings.

Also, Remove the volumeContext - `RequiredBytes` probe from the CreateVolume and use Volume's totalCapacity instead

### DIFF
--- a/pkg/node/publish_unpublish.go
+++ b/pkg/node/publish_unpublish.go
@@ -18,6 +18,7 @@ package node
 
 import (
 	"context"
+	"strings"
 
 	directv1alpha1 "github.com/minio/direct-csi/pkg/apis/direct.csi.min.io/v1alpha1"
 	"github.com/minio/direct-csi/pkg/sys"
@@ -29,6 +30,10 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+)
+
+const (
+	directCSIVolumeContextPrefix = "csi.storage.k8s.io"
 )
 
 func (n *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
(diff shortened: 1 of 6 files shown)


pkg/node/publish_unpublish.go
```diff
@@ -78,6 +84,17 @@ func (n *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
 	}
 	vol.Status.ContainerPath = containerPath
 
+	volumeContext, volumeLabels := req.GetVolumeContext(), vol.ObjectMeta.GetLabels()
+	if volumeLabels == nil {
+		volumeLabels = make(map[string]string)
+	}
+	for k, v := range volumeContext {
+		if strings.Contains(k, directCSIVolumeContextPrefix) {
```
Praveenrajmani, line 92: @wlan0  Will there be any security vulnerabilities if we add all the keys under this particular domain `csi.storage.k8s.io`? Or we can selectively add only `pod.name` and `pod.namespace` for now?

wlan0: The prefix we want is `direct.csi.min.io`. Not `csi.storage.k8s.io`

We shouldn't add `pod.name` or `pod.namespace`. We should only add the labels that comes in from the podSpec. Adding podName and podNamespace could be a security vulnerability. However, just adding the info under `direct.csi.min.io/*` labels is not.

pkg/utils/installer/rbac.go
```diff
@@ -296,6 +296,20 @@ func createClusterRole(ctx context.Context, identity string) error {
 					"direct.csi.min.io",
 				},
 			},
+			{
+				Verbs: []string{
+					clusterRoleVerbGet,
+					clusterRoleVerbList,
+					clusterRoleVerbWatch,
+				},
+				Resources: []string{
```
wlan0, line 305: 👍🏾

pkg/node/publish_unpublish.go
```diff
@@ -64,6 +94,38 @@ func (n *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublish
...
+			if strings.HasPrefix(k, directv1alpha1.Group+"/") {
+				volumeLabels[k] = v
+			}
+		}
+
+		return volumeLabels, nil
+	}
+
+	volLabels, vErr := extractPodLabels()
+	if vErr != nil {
```
wlan0, line 124: do not throw error if podInfoOnMount is not set. i.e. if pod name and namespace are missing

wlan0 (approved):

PR #154, opened 2021-01-27
Upgrade from v1alpha1 to v1beta1

Praveenrajmani:
- **Deploy conversion webhook for the upgrades and downgrades**

      (*) Add a deployment in the direct-csi-min-io namespaces which will
          expose a handler which does the conversion.
      (*) Conversion rules are added to the handler.
      (*) During the direct-csi installation, sync the CRD configs with the
          latest one
      (*) Apply the webhook settings of the conversion webhook to the CRD config
    
- **Introduce two flags to the direct-csi binary**
    
      (*) '--conversion-webhook' to start and serve conversion webhook
      (*) '--conversion-webhook-url' to pass the conversion webhook url
          (service url of the conversion deployment)
    
- **Changes to plugin flags**
    
      (*) Remove the hidden flags `--unfinalize` and `--dangerous`. Replace
          them with `--force`. CRD will not be unregistered unless the resources
          are cleaned up
      (*) Deprecate the `--crd` flag while installing. CRD will be checked and
          registered/synced by default


cmd/kubectl-direct_csi/register.go
```diff
@@ -44,28 +54,145 @@ func registerCRDs(ctx context.Context) error {
...
+			return err
+		}
+
+		existingCRD, err := crdClient.Get(ctx, crdObj.Name, metav1.GetOptions{})
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return err
+			}
+			if crdObj.Name == driveCRDName {
+				if err := setConversionWebhook(&crdObj, identity); err != nil {
```
wlan0, line 76: We should eventually make this universal, rather than per CRD type. As in, all resources should be either v1alpha1 or v1beta1 for a given release. We'll also update directcsivolumes and remove them if condition before making the next release

wlan0 (approved):

PR #158, opened 2021-01-28
make the plugin build on all three platforms - linux, windows and darwin

wlan0:
@Praveenrajmani @harshavardhana


ngbrown:
On Linux, `go` needed to be installed for the plugin to run. Was this intended? Will that be true for other platforms?

wlan0:
> On Linux, go needed to be installed for the plugin to run. Was this intended? Will that be true for other platforms?

@ngbrown we were compiling in the CRDs to get the benefits of the type system for CRD versioning. However, that was using the go binary to parse the type. A fix is in the works where we wont need the go binary anymore.

harshavardhana:
@wlan0 can you please fix the conflicts?

wlan0:
@ngbrown the PR has code that runs the CRD installer without any environment dependencies.  We missed the bug earlier because we all have go installed on our machines and minio/direct-csi in the gopath. 

It is fixed now

harshavardhana (approved):

ngbrown:
@wlan0 About dynamically rendering CRDs, I'd like to point to a document from Helm that warns against certain updates to CRDs and especially deleting them on Kubernetes.

https://github.com/helm/community/blob/7531c3c59be0472188fae9f967652c9ff9324f61/architecture/crds.md#deleting-crds

wlan0:
Yeah I'm aware, that is why we compile in the CRD version into the plugin. It is not dynamically rendered btw. It is strictly tied to the version of direct-csi it is installing. 

Helm has other issues - they have two concerns mixed into one step - templatizing and managing the deployed resources in the cluster . Helm should have never started by templatizing the YAMLs for the sake of  configurability. They should've gone the CRD way and used CRDs as the way to configure deployments in a declarative way. Helm and CRDs ended up going in opposite directions. I'm not surprised they have so many issues with CRDs.

PR #176, opened 2021-02-16
Fix ProbeMountInfo() to return correct device names

Praveenrajmani:
- The `DevName` should not have `-part-` infix
- Add test cases for coverage

Steps to test :
- Mount the `/var/lib/direct-csi/devices/XX` and then restart the drive discovery
- Mountpoint should be set correctly

Note: Before the fix, the DevName was showing up as `/var/lib/direct-csi/devices/xvdb-part-` instead of `/var/lib/direct-csi/devices/xvdb` on the mount discovery results

### DIFF
--- a/pkg/sys/sys_utils.go
+++ b/pkg/sys/sys_utils.go
@@ -67,6 +67,13 @@ func splitDevAndPartNum(s string) (string, int) {
 	possibleNum := strings.Builder{}
 	toRet := strings.Builder{}
 
+	cleanPartitionInfix := func(str string) string {
+		if strings.HasSuffix(str, directCSIPartitionInfix) {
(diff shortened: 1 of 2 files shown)


pkg/sys/sys_utils.go
```diff
@@ -67,6 +67,13 @@ func splitDevAndPartNum(s string) (string, int) {
 	possibleNum := strings.Builder{}
 	toRet := strings.Builder{}
 
+	cleanPartitionInfix := func(str string) string {
```
wlan0, line 70: This situation should never occur, right?

wlan0 (changes requested):

Praveenrajmani:
Closing this PR as we decided an alternative approach to collect mounts by (major, minor) pairs.

PR #189, opened 2021-03-17
add security checklist and seccomp profile

wlan0:
@Alevsk @Praveenrajmani


Alevsk (dismissed):
how did you get the complete list of syscalls? is strace the right tool? would like to try it locally too

wlan0:
Here is what I've updated from the default docker seccomp profile (https://github.com/moby/moby/blob/master/profiles/seccomp/default.json):

- removed all architectures other than x86_32,x86_64,amd64
- removed all ptrace syscalls
- removed bpf syscall
- removed reboot syscall
- removed ability to init, load or unload any kernel modules
- removed all raw io capabilities
- removed ability to set or manipulate system time
- removed ability to work with TTYs
- removed ability to get/set/update NICE values 
-

Alevsk (approved):
tested with `docker run -itd --security-opt seccomp:seccomp.json --security-opt apparmor:unconfined --cap-add CAP_SYS_ADMIN ubuntu bash`

PR #192, opened 2021-03-23
allow flexible org, registry and direct-csi image names

wlan0:


cmd/kubectl-direct_csi/validate_image.go
```diff
@@ -0,0 +1,2849 @@
+// This file is part of MinIO Direct CSI
+// Copyright (c) 2020 MinIO, Inc.
```
Praveenrajmani, line 2:
```suggestion
// Copyright (c) 2021 MinIO, Inc.
```

cmd/kubectl-direct_csi/validators.go
```diff
@@ -0,0 +1,121 @@
+// This file is part of MinIO Direct CSI
+// Copyright (c) 2020 MinIO, Inc.
```
Praveenrajmani, line 2:
```suggestion
// Copyright (c) 2021 MinIO, Inc.
```

cmd/kubectl-direct_csi/install.go
```diff
@@ -42,21 +43,33 @@ var (
 	installCRD       = false
 	overwriteCRD     = false
 	admissionControl = false
-	image            = "minio/direct-csi:" + Version
+	image            = "direct-csi:" + Version
```
Praveenrajmani, line 46: `quay.io/minio/direct-csi:` seems to be an invalid image (which will be chosen if `--image` is not specified) 

Can we change this to 

```
image = func() string {
     if Version == "" {
             return "direct-csi" // 'latest' tag will be chose if unspecified
     }
     return "direct-csi:" + Version
}()
 ```

wlan0: This was always the case. This will make sure they don't use dev versions of the project. Otherwise, `latest` can lead to silent failures and can be more messy than just dealing with the explicit failure during deploy.

Praveenrajmani (approved):

PR #208, opened 2021-04-15
refactor installer and generate human-friendly service names

wlan0:


cmd/kubectl-direct_csi/install.go
```diff
@@ -87,7 +87,76 @@ func install(ctx context.Context, args []string) error {
...
+				return err
+			}
+		}
+
+		labels := installer.NewObjMeta(identity, identity, "component", "controller").Labels
+		if err := installer.CreateAdmissionWebhookControllerService(ctx, labels, identity, dryRun); err != nil {
+			return err
+		}
+
+		if err := installer.RegisterAdmissionWebhookValidationRules(ctx, identity, caCertBytes, dryRun); err != nil {
```
Praveenrajmani, line 112: I think the webhook endpoint/service (controller) has to be "live" before registering the webhook config here. Otherwise, the webhook config would be treated as invalid.

pkg/utils/installer/driver-deployment.go
```diff
@@ -0,0 +1,180 @@
...
+						},
+					},
+					{
+						Name:  endpointEnvVarCSI,
+						Value: "unix:///csi/csi.sock",
+					},
+				},
+				VolumeMounts: []corev1.VolumeMount{
+					newVolumeMount(volumeNameSocketDir, "/csi", false),
+					newVolumeMount(admissionControllerCertsDir, certsDir, false),
```
Praveenrajmani, line 136: Can we mount the admission controller certs only when `admission-control` flag is enabled?

PR #209, opened 2021-04-15
Expose a metrics endpoint to export volume stats

Praveenrajmani:
Export Volume stats as metrics. DirectCSI node server exposes a metrics endpoint `/direct-csi/metrics` which will export the following set of metrics 

```
- directcsi_stats_bytes_used
- directcsi_stats_bytes_total
```

This PR also adds documentation which has a sample prometheus config and few guidelines for setting up prometheus service


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	lc := net.ListenConfig{}
+	listener, lErr := lc.Listen(ctx, "tcp", fmt.Sprintf(":%v", port))
+	if lErr != nil {
+		panic(lErr)
+	}
+
+	glog.Infof("Starting metrics exporter in port: %s", port)
+	if err := server.Serve(listener); err != nil {
+		glog.Errorf("Failed to listen and serve admission webhook server: %v", err)
+		panic(err)
```
Praveenrajmani, line 49: panic only when `err != http.ErrServerClosed`

wlan0 (approved):

PR #218, opened 2021-04-22
restrict xfs managed tree recursion to 0

wlan0:
- This can be done because when we provision new volumes, we always start with a empty directory (i.e. there are no files to traverse through)
 - When reusing old volumes, it is fair to expect that all the files already have the correct project id (i.e. no files were present before creating a volume)

@harshavardhana  @Praveenrajmani


Praveenrajmani (approved):
LGTM

PR #229, opened 2021-05-11
Handle drive-order changes on node restarts

Praveenrajmani:
- Add possible identitied to the drive like SerialNumber, PartitionUUID, FS-UUID. Which will be useful match the drives on
  restarts and order-changes.
- Moved to FS-UUID based blockdevice names and mounting
- Introduces an additional DriveStatus - `Identified` which will be set on the drive if it was unidentified
- Moved discovery as a separate module.


Praveenrajmani:
Have to fix the test-cases

wlan0 (dismissed):

wlan0 (approved):

PR #249, opened 2021-05-25
Mount drives if not mounted at new node server creation.

balamurugana:


pkg/node/node.go
```diff
@@ -72,7 +72,32 @@ func NewNodeServer(ctx context.Context, identity, nodeID, rack, zone, region str
 	topologies[topology.TopologyDriverNode] = nodeID
 	driveTopology := topologies
 
+	mounts, err := sys.ProbeMountInfo()
+	if err != nil {
+		return nil, err
+	}
+
 	for _, drive := range drives {
+		isMounted := false
```
wlan0, line 81: Just call sys.SafeMount while passing in prjquota

wlan0 (approved):

PR #257, opened 2021-06-01
Replace glog with klog.

balamurugana:

### DIFF
--- a/cmd/direct-csi/cmd.go
+++ b/cmd/direct-csi/cmd.go
@@ -25,7 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
-	_ "github.com/golang/glog"
+	"k8s.io/klog"
 )
 
 var Version string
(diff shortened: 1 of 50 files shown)


cmd/direct-csi/cmd.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
-	_ "github.com/golang/glog"
+	_ "k8s.io/klog"
```
wlan0, line 28: if you set --v, it wont understand the argument. You need to initialize flags like its done here: https://github.com/kubernetes-sigs/container-object-storage-interface-controller/blob/master/cmd/controller-manager/controller-manager.go#L37

wlan0:
LGTM, expect one comment

cmd/direct-csi/cmd.go
balamurugana: Done

wlan0 (approved):

PR #261, opened 2021-06-03
Verify drive mounts on restarts

Praveenrajmani:
- Mount drive to its respective Mountpoint

Steps to test :-

- Reboot a node
- Check the mounts are present after the direct-csi starts by looking into `df -h` and `lsblk` / `cat /proc/1/mountinfo`


pkg/utils/installer/install.go
```diff
@@ -366,7 +366,6 @@ func CreateDaemonSet(ctx context.Context, identity string, directCSIContainerIma
 				Name:  nodeDriverRegistrarContainerName,
 				Image: filepath.Join(registry, org, nodeDriverRegistrarContainerImage),
 				Args: []string{
-					fmt.Sprintf("--v=%d", logLevel),
```
Praveenrajmani, line 369: Removing this because, currently, the container fails to start with of the following error

```
Error: unknown flag: --v
```

We can address this in a separate PR

wlan0: This should not be removed. This will make us backwards incompatible

wlan0 (approved):

PR #262, opened 2021-06-03
Set LogLevel to 0 for logs in install/uninstall

Praveenrajmani:
Right now, the installation and uninstallation does not print any log lines

```
➜  direct-csi git:(fix-klog) ./kubectl-direct_csi install --org praveen0 --image direct-csi:remounts
➜  direct-csi git:(fix-klog
```

### DIFF
--- a/cmd/kubectl-direct_csi/install.go
+++ b/cmd/kubectl-direct_csi/install.go
@@ -95,14 +95,14 @@ func install(ctx context.Context, args []string) error {
 		}
 	}
 	if !dryRun {
-		klog.V(1).Infof("'%s' namespace created", utils.Bold(identity))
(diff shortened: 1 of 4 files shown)


cmd/kubectl-direct_csi/install.go
```diff
@@ -95,14 +95,14 @@ func install(ctx context.Context, args []string) error {
 		}
 	}
 	if !dryRun {
-		klog.V(1).Infof("'%s' namespace created", utils.Bold(identity))
```
wlan0, line 98: Dont specify a V level if it is 0

wlan0: just `klog.Infof ` will do

wlan0 (approved):

PR #263, opened 2021-06-07
Create Staging target path if it does not exist while staging

Praveenrajmani:
Do not fail stage volume calls if the target path is not present. setting `xfs_quota` might fail with following error in such `DoesNotExist` scenarios

```
MountVolume.MountDevice failed for volume "pvc-4c0a8e78-8450-494d-87e5-daf656485e96" : rpc error: code = Internal desc = failed stage volume: rpc error: code = Internal desc = Error while setting xfs limits: SetQuota failed for pvc-4c0a8e78-8450-494d-87e5-daf656485e96 with error: (exit status 1), output: (xfs_quota: cannot setup path for project dir /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-4c0a8e78-8450-494d-87e5-daf656485e96/globalmount: No such file or directory )
```

### DIFF
--- a/pkg/node/stage_unstage.go
+++ b/pkg/node/stage_unstage.go
@@ -68,6 +68,10 @@ func (n *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolu
 		return nil, err
 	}
 
+	if err := os.MkdirAll(stagingTargetPath, 0755); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


Praveenrajmani:
Similar fix was done in NodePublishVolume earlier. For reference - https://github.com/minio/direct-csi/pull/181

balamurugana (approved):

PR #270, opened 2021-06-09
Make flat structure in device probing.

balamurugana:


pkg/sys/drive_discovery_linux.go
```diff
@@ -21,56 +21,56 @@ import (
...
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"syscall"
 
 	"golang.org/x/sys/unix"
-	"k8s.io/klog/v2"
+	"k8s.io/klog"
```
wlan0, line 32: this should be klog/v2

pkg/node/discovery/drive_matcher.go
```diff
@@ -100,7 +98,7 @@ func (d *Discovery) identifyDriveByLegacyName(localDriveState directcsi.DirectCS
 	}
 
 	for i, remoteDrive := range d.remoteDrives {
-		if !remoteDrive.matched && remoteDrive.Name == v1beta1DriveName {
+		if !remoteDrive.matched && remoteDrive.Status.NodeName == d.NodeID && remoteDrive.Status.Path == localDriveState.Path {
```
Praveenrajmani, line 101: We don't need this condition - `remoteDrive.Status.NodeName == d.NodeID` as the `remoteDrives` are already filtered by NodeName

pkg/utils/utils.go
```diff
@@ -78,3 +80,28 @@ func DefaultIfZeroFloat(left, right float32) float32 {
...
+		name := strings.ReplaceAll(
+			strings.Replace(devName, sys.DirectCSIPartitionInfix, "", 1),
+			sys.DirectCSIPartitionInfix,
+			sys.HostPartitionInfix,
+		)
+		return filepath.Join(sys.HostDevRoot, name)
+	}
+}
+
+func GetDrivePath(drive *directcsi.DirectCSIDrive) string {
```
Praveenrajmani, line 100: We don't need to switch case here on versions, because, there can be v1beta2 drives as well with the `Path` set to `/var/lib/direct-csi`  notation

Any fresh installations from v1.4.0 also will have `/var/lib/direct-csi` notation on `Path`

We can directly use getRootBlockFile wherever needed

balamurugana: Done

pkg/sys/drive_discovery_linux.go
balamurugana: Done

pkg/node/discovery/drive_matcher.go
balamurugana: Done

pkg/node/discovery/utils.go
```diff
@@ -35,24 +36,30 @@ func (d *Discovery) verifyDriveMount(existingDrive *directcsi.DirectCSIDrive) er
...
-				isMounted = true
-				break
+
+		major, minor, err := sys.GetMajorMinor(mountSource)
+		if err != nil {
+			return err
+		}
+
+		mounted := false
+		if mounts, found := d.mounts[fmt.Sprintf("%v:%v", major, minor)]; found {
```
wlan0, line 52: hide the string construction behind a struct method

```
d.getMountInfo(major, minor int) []sys.MountInfo
```

pkg/node/discovery/discovery_test.go
```diff
@@ -0,0 +1,47 @@
+// This file is part of MinIO Direct CSI
```
Praveenrajmani, line 1: we can remove this file as it is not used

pkg/node/discovery/utils.go
balamurugana: Done

pkg/node/discovery/discovery_test.go
balamurugana: Done

wlan0 (approved):

wlan0 (approved):

PR #291, opened 2021-07-06
Move to UBI containers

Praveenrajmani:
- Uses RHEL7 compatible ubi7-minimal image for Dockerfile.rhel
- Use ubi8-minimal as default
- Using centos' xfsprogs packages for compatibility


Dockerfile
```diff
@@ -1,11 +1,17 @@
...
 
-RUN apk add -U --no-cache ca-certificates xfsprogs xfsprogs-extra
-
 COPY direct-csi /direct-csi
 COPY CREDITS /licenses/CREDITS
 COPY LICENSE /licenses/LICENSE
 
-ENTRYPOINT ["/direct-csi"]
+COPY centos7.repo /etc/yum.repos.d/centos7.repo
+COPY RPM-GPG-KEY-CentOS-7 /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
```
Praveenrajmani, line 10: Instead of COPY-ing, we can actually use curl and redirect it to `/etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7`

Dockerfile
```diff
@@ -1,11 +1,17 @@
-FROM alpine:latest
+FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
```
harshavardhana, line 1: Do not use latest always fix a tag.

harshavardhana (changes requested):

Dockerfile
balamurugana: The idea here is to stick to use the latest Red Hat UBI 7 i.e. it won't pull other RHEL specified version.

Praveenrajmani:
Any idea why GoReleaser is failing for `arm64` build? @harshavardhana @wlan0 @balamurugana

Dockerfile
harshavardhana: We should never pull latest, that is how containers are built. We should always stick to a tag.

Dockerfile
```diff
@@ -1,11 +1,17 @@
...
 COPY direct-csi /direct-csi
 COPY CREDITS /licenses/CREDITS
 COPY LICENSE /licenses/LICENSE
 
-ENTRYPOINT ["/direct-csi"]
+COPY centos7.repo /etc/yum.repos.d/centos7.repo
+
+RUN \
+    curl https://www.centos.org/keys/RPM-GPG-KEY-CentOS-7 -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    mv /etc/yum.repos.d/ubi.repo /etc/yum.repos.d/ubi.repo.old  && \
```
harshavardhana, line 13: Renaming ubi.repo is not correct enable relevant repos such that we can also get security fixes for other packages, disabling this means that future updates to packages are disabled.

harshavardhana (changes requested):

.goreleaser.yml
```diff
@@ -25,7 +25,6 @@ builds:
       - amd64
       - arm64
       - ppc64le
-      - s390x
```
Praveenrajmani, line 28: Removed `s390x` because, it is neither supported by RHEL7 nor RHEL8

.goreleaser.yml
```diff
@@ -77,38 +75,9 @@ dockers:
   extra_files:
     - LICENSE
     - CREDITS
+    - centos7.repo
   build_flag_templates:
   - "--platform=linux/amd64"
-- image_templates:
```
Praveenrajmani, line 82: RHEL7 only supports `amd64`. So, removing the other archs from the run.

Dockerfile.dev
```diff
@@ -0,0 +1,17 @@
...
+COPY direct-csi /direct-csi
+COPY CREDITS /licenses/CREDITS
+COPY LICENSE /licenses/LICENSE
+
+COPY centos.repo /etc/yum.repos.d/centos.repo
+
+RUN \
+    curl https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official && \
+    mv /etc/yum.repos.d/ubi.repo /etc/yum.repos.d/ubi.repo.old && \
+    microdnf install xfsprogs --nodocs && \
```
harshavardhana, line 14: You should also do `microdnf update` as well to get latest updates if any?

harshavardhana (changes requested):

wlan0 (dismissed):

Dockerfile.dev
wlan0: It looks like centos is the upstream now, and microdnf update does not always give signed packages. It seems liek the best approach for security reasons is we should use specific package versions when updating.

harshavardhana: I meant we need to update for RedHat packages, and perhaps install xfsprogs and then remove the CentOS.repo just in case to avoid any additional repos in the container.

harshavardhana (dismissed):

harshavardhana (approved):

PR #294, opened 2021-07-07
Add PodSecurityPolicy creation and binding.

### LINKED ISSUE #292: Automatically install a podsecurity policy during direct-csi install

balamurugana:
Fixes #292

### DIFF
--- a/pkg/utils/installer/install.go
+++ b/pkg/utils/installer/install.go
@@ -20,11 +20,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"k8s.io/klog"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
+
+	"k8s.io/klog/v2"
 
 	"github.com/minio/direct-csi/pkg/topology"
 	"github.com/minio/direct-csi/pkg/utils"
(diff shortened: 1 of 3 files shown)


wlan0 (approved):

wlan0:
LGTM

pkg/utils/installer/install.go
```diff
@@ -431,6 +431,10 @@ func CreateDaemonSet(ctx context.Context, identity string, directCSIContainerIma
 				}(),
 				SecurityContext: &corev1.SecurityContext{
 					Privileged: &privileged,
+					SeccompProfile: &corev1.SeccompProfile{
```
Praveenrajmani, line 434: Should we enable Seccomprofile by default?

pkg/utils/installer/install.go
```diff
@@ -509,6 +513,7 @@ func CreateDaemonSet(ctx context.Context, identity string, directCSIContainerIma
 					Namespace: sanitizeName(name),
 					Annotations: map[string]string{
 						CreatedByLabel: DirectCSIPluginName,
+						"container.apparmor.security.beta.kubernetes.io/direct-csi": apparmorProfileName,
```
Praveenrajmani, line 516: Should we enable apparmour by default?

Praveenrajmani (commented):
Can we maintain flags for seccomp, apparmour and selinux? So that, only when we any of these flags,  we can go and enable it in the containers.

pkg/utils/installer/install.go
balamurugana: No. Made it optional now.

pkg/utils/installer/install.go
balamurugana: No. Made it optional now.

wlan0 (approved):

PR #312, opened 2021-07-30
Remove go.mongodb.org/mongo-driver from go.mod

Praveenrajmani:
the project doesn't use `go.mongodb.org/mongo-driver`, So, removing it from go.mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,6 @@ require (
 	github.com/smartystreets/assertions v1.1.1 // indirect
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/viper v1.8.1
-	go.mongodb.org/mongo-driver v1.1.2 // indirect
 	golang.org/x/sys v0.0.0-20210510120138-977fb7262007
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -32,7 +32,6 @@ require (
 	github.com/smartystreets/assertions v1.1.1 // indirect
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/viper v1.8.1
-	go.mongodb.org/mongo-driver v1.1.2 // indirect
```
harshavardhana, line 35: that's an indirect dep - it's a bit clunky style of introduction not sure why `go mod` adds it. 

Although your intentions are right in the PR - the problem is it might still get again added since it's some kind of dependency management behavior. 

Let's take in and see.

harshavardhana (approved):

go.mod
wlan0: go mod why said it’s coming from spf13/viper v1.7. We upgraded to v1.8.1 in a previous PR, but go mod tidy didn’t remove this line. 

This time, go mod why says, this dependency is not needed. Even without removing the line. That’s why we are ok with manually removing it

PR #319, opened 2021-08-09
Implement setprojectid and setprojectquota in pure-go

Praveenrajmani:
This will remove the exec calls for setting the project-id and project-quota


pkg/sys/fs/quota_linux.go
```diff
@@ -0,0 +1,163 @@
...
+func (q *FSQuota) SetProjectID(projectID uint32) error {
+
+	targetDir, err := os.Open(q.GetPath())
+	if err != nil {
+		return fmt.Errorf("could not open %v: %v", q.GetPath(), err)
+	}
+	defer targetDir.Close()
+
+	var fsx FSXAttr
+	if _, _, errno := syscall.Syscall(syscall.SYS_IOCTL,
```
balamurugana, line 66: You could keep low level xfs syscalls as separate package.

wlan0 (approved):

PR #330, opened 2021-08-13
Handle abnormal delete volume calls and fix pv directory clean-up upon volume deletion 

Praveenrajmani:
- Do not delete volumes if an abnormal delete is called on directcsivolume in use.
- Fix DeleteVolume RPC to not proceed if the volume hasn't been unstaged
- Fix deletion of PV directories upon deletion (FIX: Do not unset HostPath on NodeUnstageVolume RPC call)

Fixes https://github.com/minio/direct-csi/issues/323

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -387,7 +387,13 @@ func (c *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolu
 		return nil, status.Errorf(codes.NotFound, "could not retreive volume [%s]: %v", vID, err)
 	}
 
-	if vol.Status.StagingPath != "" || vol.Status.ContainerPath != "" {
+	// Do not proceed if the volume hasn't been unpublished or unstaged
+	if utils.IsConditionStatus(vol.Status.Conditions,
+		string(directcsi.DirectCSIVolumeConditionStaged),
+		metav1.ConditionTrue) ||
+		utils.IsConditionStatus(vol.Status.Conditions,
+			string(directcsi.DirectCSIVolumeConditionPublished),
+			metav1.ConditionTrue) {
 		return nil, status.Errorf(codes.FailedPrecondition,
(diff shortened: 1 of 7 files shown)


pkg/node/stage_unstage.go
```diff
@@ -154,7 +154,6 @@ func (n *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstage
 		}
 	}
 
-	vol.Status.HostPath = ""
```
Praveenrajmani, line 157: As we were unset-ing the HostPath here, `os.RemoveAll(hostPath)` on deletion becomes in-effective. 

This fix would clean up the host path on volume deletion.

pkg/controller/controller.go
```diff
@@ -387,6 +387,11 @@ func (c *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolu
 		return nil, status.Errorf(codes.NotFound, "could not retreive volume [%s]: %v", vID, err)
 	}
 
+	if vol.Status.StagingPath != "" || vol.Status.ContainerPath != "" {
```
Praveenrajmani, line 390: Do not proceed deletion if the volume hasn't been unpublished and unstaged. Repond with proper error code (codes.FailedPrecondition) if the DeleteVolume RPC was prematurely called.

wlan0: It is better to check status instead of weak signals such as StagingPath and ContainerPath for staging and publishing.

pkg/node/stage_unstage.go
wlan0: Nice catch.

wlan0:
LGTM. One small fix needed, then I'll merge

Praveenrajmani:
Addressed the review comment @wlan0 . PTAL.

wlan0 (approved):

PR #340, opened 2021-08-20
[WIP] [Do Not Merge] Add upgrade tests in github CI

Praveenrajmani:


.github/workflows/upgrade-test.yaml
```diff
@@ -0,0 +1,52 @@
+name: Upgrade Tests
```
balamurugana, line 1: Why can't this be part of integration test itself? If naming is the issue, change the name.

hack/ci/check-upgrade.sh
```diff
@@ -0,0 +1,151 @@
...
+function setup_lvm() {
+    loopdev=$(create_loop testpv.img 1G)
+    sudo pvcreate "$loopdev"
+    vgname="testvg$RANDOM"
+    sudo vgcreate "$vgname" "$loopdev"
+    sudo lvcreate --name=testlv --extents=100%FREE "$vgname"
+    LV_DEVICE=$(readlink -f "/dev/$vgname/testlv")
+}
+
+function setup_luks() {
```
balamurugana, line 53: Do not duplicate code instead reuse.

hack/ci/check-upgrade.sh
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env bash
```
balamurugana, line 1: Why do we need a separate script and why can't be part of `check.sh` itself?

hack/ci/check-upgrade.sh
```diff
@@ -0,0 +1,151 @@
...
+
+function setup_luks() {
+    loopdev=$(create_loop testluks.img 1G)
+    echo "mylukspassword" > lukspassfile
+    yes YES | sudo cryptsetup -y -v luksFormat "$loopdev" lukspassfile
+    sudo cryptsetup -v luksOpen "$loopdev" myluks --key-file=lukspassfile
+    LUKS_DEVICE=$(readlink -f /dev/mapper/myluks)
+}
+
+function install_directcsi_v1.3.6() {
```
balamurugana, line 61: As installation is generic for current and previous releases, have a version independent function for install/uninstall.

hack/ci/lvm-setup.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
balamurugana, line 1: Why is this needed?

Praveenrajmani: This is just a WIP PR, Will need to do some changes @balamurugana

PR #345, opened 2021-08-24
Fix returning error on drive update handler

Praveenrajmani:
The `err` variable was being overridden (masked) while updating the drive


balamurugana (approved):
LGTM

wlan0 (approved):

PR #347, opened 2021-08-24
Only include direct-csi formatted drives (Ready|InUse drives) in the info output

Praveenrajmani:


balamurugana:
PR https://github.com/minio/direct-csi/pull/344 is already there for the same fix.

PR #359, opened 2021-08-28
Enhance functional tests previously called integration tests

balamurugana:


functests/common.sh
```diff
@@ -0,0 +1,161 @@
...
+        fi
+    done
+}
+
+function _uninstall_directcsi() {
+    "${DIRECT_CSI_CLIENT}" uninstall "$@"
+
+    pending=7
+    while [[ $pending -gt 0 ]]; do
+        echo "$ME: waiting for direct-csi pods to go down... ${pending}"
```
Praveenrajmani, line 76: we wont delete the conversion webhook pods while uninstalling with v1.3.6. We implemented that change in v1.4.0 where we bring down conversion webhook deployment during uninstallation. 

(cancelling the CI workflow as the script waits for this condition to be satisfied)

wlan0 (approved):

PR #369, opened 2021-08-31
Support `--all` flag in volumes listing

### LINKED ISSUE #367: Support for `--all` flag in volumes listing
Currently, We list only `Ready` volumes on `volumes ls` command which may skip the non-provisioned (or) yet-to-be provisioned volumes.

We can have an `--all` flag on `volumes ls` command, which may list all volumes irrespective of their provisioning state. 

cc @wlan0

Praveenrajmani:
Fixes #367

### DIFF
--- a/cmd/kubectl-direct_csi/volumes_list.go
+++ b/cmd/kubectl-direct_csi/volumes_list.go
@@ -78,6 +78,7 @@ func init() {
 	listVolumesCmd.PersistentFlags().StringSliceVarP(&accessTiers, "access-tier", "", accessTiers, "filter based on access-tier")
 	listVolumesCmd.PersistentFlags().StringSliceVarP(&podNames, "pod-name", "", podNames, "glob prefix match for pod names")
 	listVolumesCmd.PersistentFlags().StringSliceVarP(&podNss, "pod-namespace", "", podNss, "glob prefix match for pod namespace")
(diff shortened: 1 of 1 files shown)


cmd/kubectl-direct_csi/volumes_list.go
```diff
@@ -78,6 +78,7 @@ func init() {
 	listVolumesCmd.PersistentFlags().StringSliceVarP(&accessTiers, "access-tier", "", accessTiers, "filter based on access-tier")
 	listVolumesCmd.PersistentFlags().StringSliceVarP(&podNames, "pod-name", "", podNames, "glob prefix match for pod names")
 	listVolumesCmd.PersistentFlags().StringSliceVarP(&podNss, "pod-namespace", "", podNss, "glob prefix match for pod namespace")
+	listVolumesCmd.PersistentFlags().BoolVarP(&all, "all", "a", all, "list all volumes (including non-provisioned)")
```
balamurugana, line 81: if short flag `a` is not used to denote `all` already, we could use `A` like `kubectl get pods` command.

balamurugana (approved):
LGTM

balamurugana (changes requested):
Please revert to short flag `a` for consistent with `drives list`

wlan0:
merging this since @balamurugana's comments have been addressed.

PR #379, opened 2021-09-15
Add udev event handling

balamurugana:
Depends on https://github.com/minio/direct-csi/pull/378


Praveenrajmani:
Can we enable `--dynamic-discovery` by default @balamurugana ?

Also, the security policies - seccomp and apparmour profiles requires an update right? As we brought in new volume mounts for dynamic discovery?

balamurugana:
> Can we enable `--dynamic-discovery` by default @balamurugana ?
> 
> Also, the security policies - seccomp and apparmour profiles requires an update right? As we brought in new volume mounts for dynamic discovery?

Updated `apparmor.profile` and no change in `seccomp.json`

PR #380, opened 2021-09-20
Do not display error message for released drives while listing

### LINKED ISSUE #314: Do not display error conditions for Released drives
Right now, we don't clear the error conditions when the drive is released. We should clear them out / skip displaying errors for released drives. This avoids unnecessary display of errors on released drives.

sinhaashish:
Details in issue https://github.com/minio/direct-csi/issues/314
Closes #314

### DIFF
--- a/cmd/kubectl-direct_csi/drives_list.go
+++ b/cmd/kubectl-direct_csi/drives_list.go
@@ -167,6 +167,10 @@ func listDrives(ctx context.Context, args []string) error {
 
 		msg := ""
 		for _, c := range d.Status.Conditions {
+			if d.Status.DriveStatus == directcsi.DriveStatusReleased {
+				// Do not diplay error in case of released drives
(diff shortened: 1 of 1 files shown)


cmd/kubectl-direct_csi/drives_list.go
```diff
@@ -167,6 +167,10 @@ func listDrives(ctx context.Context, args []string) error {
 
 		msg := ""
 		for _, c := range d.Status.Conditions {
+			if string(d.Status.DriveStatus) == string(directcsi.DriveStatusReleased) {
```
Praveenrajmani, line 170: We don't need to convert to `string`,  we can directly check the field value. 
```suggestion
			if d.Status.DriveStatus == directcsi.DriveStatusReleased {
```

Praveenrajmani (approved):

PR #387, opened 2021-09-23
Ellipses expansion

sinhaashish:


pkg/utils/rangeexpander/rangeexpander.go
```diff
@@ -0,0 +1,250 @@
+package rangeexpander
```
balamurugana, line 1: Below logic is much simpler for ellipses support. For further usage, you would need to add permutation expansion in this code.
```go
import (
	"fmt"
	"os"
	"regexp"
	"strconv"
	"strings"
)

var alphaRegexp = regexp.MustCompile("^[a-z]+$")

func alpha2int(value string) (ui64 uint64) {
	p := uint64(1)
	for i := len(value) - 1; i >= 0; i-- {
		ui64 += uint64(value[i]-96) * p
		p *= 26
	}
	return ui64
}

func int2alpha(ui64 uint64) (value string) {
	for {
		r := ui64 % 26
		if r == 0 {
			r = 26
		}
		value = string(byte(r+96)) + value

		if ui64 <= 26 {
			break
		}
		ui64 /= 26
	}

	return value
}

type ellipsis struct {
	start   uint64
	end     uint64
	isAlpha bool

	startIndex int
	endIndex   int
}

func (e *ellipsis) expand() (values []string) {
	start, end := e.start, e.end
	if start > end {
		start, end = end, start
	}

	for i := start; i <= end; i++ {
		if e.isAlpha {
			values = append(values, int2alpha(i))
		} else {
			values = append(values, fmt.Sprintf("%v", i))
		}
	}

	return values
}

func parseEllipsis(arg string, start, end int) (*ellipsis, error) {
	pattern := arg[start:end]

	parseValue := func(value string) (ui64 uint64, isAlpha bool, err error) {
		if ui64, err = strconv.ParseUint(value, 10, 64); err == nil {
			return ui64, false, nil
		}

		if alphaRegexp.MatchString(value) {
			return alpha2int(value), true, nil
		}

		return 0, false, err
	}

	tokens := strings.Split(arg[start+1:end-1], "...")
	switch len(tokens) {
	case 0, 1:
		return nil, fmt.Errorf("%v: invalid ellipsis %v at %v", arg, pattern, start)
	}

	startValue, isAlphaStart, err := parseValue(tokens[0])
	if err != nil {
		return nil, fmt.Errorf("%v: invalid start value '%v' in ellipsis %v at %v", arg, tokens[0], pattern, start)
	}

	endValue, isAlphaEnd, err := parseValue(tokens[1])
	if err != nil {
		return nil, fmt.Errorf("%v: invalid end value '%v' in ellipsis %v at %v", arg, tokens[1], pattern, start)
	}

	if isAlphaStart != isAlphaEnd {
		return nil, fmt.Errorf("%v: invalid ellipsis %v at %v; start/end must be same kind", arg, pattern, start)
	}

	return &ellipsis{
		start:      startValue,
		end:        endValue,
		isAlpha:    isAlphaStart,
		startIndex: start,
		endIndex:   end,
	}, nil
}

func getEllipses(arg string) (ellipses []*ellipsis, err error) {
	curlyOpened := false
	start := 0
	for i, c := range arg {
		switch c {
		case '{':
			if curlyOpened {
				return nil, fmt.Errorf("%v: nested ellipsis pattern at %v", arg, i+1)
			}

			curlyOpened = true
			start = i

		case '}':
			if !curlyOpened {
				return nil, fmt.Errorf("%v: invalid ellipsis pattern at %v", arg, i+1)
			}
			curlyOpened = false

			ellipsis, err := parseEllipsis(arg, start, i+1)
			if err != nil {
				return nil, err
			}

			ellipses = append(ellipses, ellipsis)
		}
	}

	return ellipses, nil
}

func main() {
	if len(os.Args) != 2 {
		fmt.Printf("usage: %v <VALUE>\n", os.Args[0])
		os.Exit(-1)
	}

	arg := os.Args[1]

	ellipses, err := getEllipses(arg)
	if err != nil {
		panic(err)
	}

	for i := range ellipses {
		fmt.Printf("%+v\n", ellipses[i])
		fmt.Println(ellipses[i].expand())
	}
}
```

sinhaashish: Thanks @balamurugana , checking this

pkg/utils/rangeexpander/rangeexpander_test.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		return 0
+	}
+	testCases := []struct {
+		arg         string
+		ellipses    []*ellipsis
+		errReturned bool
+	}{
+		// Valid case
+		{"{a...z}", []*ellipsis{{start: parseValue("a"), end: parseValue("z"), isAlpha: true, startIndex: 0, endIndex: 7}}, false},
```
wlan0, line 93: {123}
{a...{a...z}}
{a...}z
{1.{...{zz}
1}ccc{sss}

wlan0: {11...az}

pkg/utils/rangeexpander/rangeexpander_test.go
```diff
@@ -0,0 +1,138 @@
...
+		{"{a...c}p{0...2}", []string{"ap0", "ap1", "ap2", "bp0", "bp1", "bp2", "cp0", "cp1", "cp2"}, false},
+		// Valid case- Start with ellipsis
+		{"{a...c}p{0...2}9", []string{"ap09", "ap19", "ap29", "bp09", "bp19", "bp29", "cp09", "cp19", "cp29"}, false},
+		// Valid case- Start with ellipsis
+		{"{a...c}p{0...2}9{d...a}", []string{"ap09a", "ap09b", "ap09c", "ap09d", "ap19a", "ap19b", "ap19c", "ap19d", "ap29a", "ap29b", "ap29c", "ap29d", "bp09a", "bp09b", "bp09c", "bp09d", "bp19a", "bp19b", "bp19c", "bp19d", "bp29a", "bp29b", "bp29c", "bp29d", "cp09a", "cp09b", "cp09c", "cp09d", "cp19a", "cp19b",
+			"cp19c", "cp19d", "cp29a", "cp29b", "cp29c", "cp29d"}, false},
+		// Valid case- Start with non-ellipsis
+		{"abc", []string{"abc"}, false},
+		// Valid case- Start with non-ellipsis
+		{"ab{p...r}", []string{"abp", "abq", "abr"}, false},
```
wlan0, line 50: a{12...120}x
ax{ab...dx}y

wlan0 (approved):

harshavardhana (commented):
It would be nice to submit this to `github.com/minio/pkg/expander`  and then use `github.com/minio/pkg` in direct-csi? 

Is direct-csi trying to limit `go.mod` imports?

wlan0:
@harshavardhana this impl is specifically for drives and node domain addresses. 

we can contribute it back to the `minio/pkg`

PR #393, opened 2021-10-01
Add context based fs probing

balamurugana:


pkg/fs/ext4/ext4.go
```diff
@@ -165,69 +166,37 @@ type EXT4SuperBlock struct {
...
-	if ext4.SuperBlock.HasExtended() {
-		return getUUID(ext4.SuperBlock.SUuid)
-	}
-
-	return strconv.Itoa(int(ext4.SuperBlock.DefaultReserveUID)) + strconv.Itoa(int(ext4.SuperBlock.DefaultReserveGID)), nil
-}
-
-func (ext4 *EXT4) Type() string {
-	return FSTypeEXT4
+	return UUID2String([16]byte{})
```
Praveenrajmani, line 174: Earlier, if it is not a dynamic revisioned block (Does not have FSUUID), we take `DefaultReserveUID+DefaultReserveGID` as ID.

If it is unique, can we use that as an ID() for dynamic revisioned ext4 drives?

pkg/fs/probe.go
```diff
@@ -0,0 +1,154 @@
...
+// FS denotes filesystem interface.
+type FS interface {
+	ID() string
+	Type() string
+	TotalCapacity() uint64
+	FreeCapacity() uint64
+}
+
+func probe(device string) (FS, error) {
+	devFile, err := os.OpenFile(device, os.O_RDONLY, os.ModeDevice)
```
Praveenrajmani, line 43: Suggestion - earlier we had the following style

```
filesystems := []Filesystem{
		xfs.NewXFS(),
		ext4.NewEXT4(),
		fat32.NewFAT32(),
		swap.NewSwap(),
		// Add new filesystems here
	}

	for _, fs := range filesystems {
		found, err := fs.ProbeFS(device, 0)
		if err != nil {
			return nil, err
		}
		if found {
			return fs, nil
		}
	}
```

Can we keep this style if possible?

pkg/fs/xfs/xfs.go
```diff
@@ -18,20 +18,25 @@ package xfs
...
-type XFS struct {
-	SuperBlock *XFSSuperBlock
-}
+	fserrors "github.com/minio/direct-csi/pkg/fs/errors"
+)
 
-func NewXFS() *XFS {
-	return &XFS{}
+// UUID2String converts UUID to string.
+func UUID2String(uuid [16]byte) string {
```
Praveenrajmani, line 28: Can we move this function to a common place and use it in ext4.go also?

pkg/fs/ext4/ext4.go
balamurugana: `fs-revision-level` i.e. `RevLevel` is set to zero means to be backward compatible with linux kernel <= 1.2. In this case, using of default reserved GID/UID does not help to get uniqueness of different devices. Returning zero and generating new UUID at usage is more appropriate.

pkg/fs/probe.go
balamurugana: No. Each filesystem probing accepts different arguments. In current code, `fat32` requires `io.ReadSeeker` whereas others require `io.Reader`. Current is working because each filesystem opens device independently. This is what fixed here.

pkg/fs/xfs/xfs.go
balamurugana: Each filesystem represents UUID data in a different format. It is OK to be duplicated this. Why this function exported is someone can directly use `xfs` or any filesystem code at low level and calls this function for UUID to string conversion if needed.

PR #397, opened 2021-10-06
Fix `kubectl direct-csi volumes list` command

Praveenrajmani:
Fix `MatchTrueConditions` to return true if no statuslist is provided


Praveenrajmani:
Also, we need to check why functional test did not catch this

PR #419, opened 2021-10-28
Remove own dev directory management

balamurugana:


pkg/drive/drive.go
```diff
@@ -18,40 +18,82 @@ package drive
...
+	}
+	return "/dev/" + name, nil
+}
+
+func mknod(fsuuid string, major, minor uint32) error {
+	path := filepath.Join(sys.DirectCSIDevRoot, fsuuid)
+
+	if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
```
Praveenrajmani, line 63: question: will `os.Remove` throw a "device busy" if the device is mounted?

Praveenrajmani: can we mknod only if the (major, minor) of the device is different from what we expect? 

i see that, we call this function during drive formatting, in-case if the drive formatting fails due to any error, the listener will retry the format requests multiple times.

in such cases, we may delete and create the device files too many times.

Praveenrajmani (approved):
Few questions on `mknod`

LGTM otherwise

pkg/drive/drive.go
balamurugana: 1. I haven't checked that. For regular files, `os.Remove()` succeeded even file is in use.
2. The better way is just do `Mknod()` and ignore `os.ErrExist` error.

wlan0 (dismissed):

pkg/drive/drive.go
```diff
@@ -18,40 +18,78 @@ package drive
...
+	name, err := sys.GetDeviceName(major, minor)
+	if err != nil {
+		return "", err
+	}
+	return "/dev/" + name, nil
+}
+
+func mknod(fsuuid string, major, minor uint32) error {
+	path := filepath.Join(sys.DirectCSIDevRoot, fsuuid)
+	err := unix.Mknod(path, unix.S_IFBLK|0666, int(unix.Mkdev(major, minor)))
```
Praveenrajmani, line 60: can we remove and create? to handle path changes? Like we had earlier

wlan0 (approved):

PR #432, opened 2021-11-09
Update drive name in drive/volume labels

### LINKED ISSUE #431: Update drive and volume object labels when drive states are synced in dynamic discovery process
We use drive and volume object labels like `path` to list volumes and drives. When drive order changes and uevent handler detects it, we need to update those labels in corresponding drive and volume CRD objects.

Ref: https://github.com/minio/direct-csi/blob/master/pkg/node/uevent.go#L73 
(If the drive order changes and are detected using udev dynamic discovery, we need to update the `path` labels in drives and corresponding volumes as well)

We anyway handle these on restarts during discovery process. we can also do the same while updating states in udev handlers.

balamurugana:
Fixes #431

### DIFF
--- a/pkg/node/uevent.go
+++ b/pkg/node/uevent.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"os"
 	"path"
+	"strings"
 	"sync"
 	"time"
 
@@ -32,6 +33,7 @@ import (
 	"github.com/minio/direct-csi/pkg/utils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/node/uevent.go
```diff
@@ -69,13 +70,35 @@ func (handler *ueventHandler) syncDrive(
...
+			if err == nil && nameChanged {
+				volumeInterface := handler.directCSIClient.DirectV1beta3().DirectCSIVolumes()
+				for _, finalizer := range drive.GetFinalizers() {
+					if !strings.HasPrefix(finalizer, directcsi.DirectCSIDriveFinalizerPrefix) {
+						continue
+					}
+
+					volumeName := strings.TrimPrefix(finalizer, directcsi.DirectCSIDriveFinalizerPrefix)
+					volume, err := volumeInterface.Get(ctx, volumeName, metav1.GetOptions{TypeMeta: utils.DirectCSIVolumeTypeMeta()})
+					if err != nil {
```
Praveenrajmani, line 91: can we retry on conflict here? There could be a scenario where drive path gets updated successfully but volume update fails due to update conflict

balamurugana: Done

Praveenrajmani (approved):

wlan0 (approved):

PR #438, opened 2021-11-12
Fix client-gen to add all the versions in clientset.go

Praveenrajmani:


hack/update-codegen.sh
```diff
@@ -78,19 +78,22 @@ for version in "${versions[@]}"; do
...
+# Join array elements with ",".
+input_versions=$(IFS=,; echo "${versions_arr[*]}")
+# client
+client-gen \
+    --go-header-file "${SCRIPT_ROOT}/boilerplate.go.txt" \
+    --input-dirs "${input_dirs}" \
+    --output-package "${REPOSITORY}/pkg/" \
+    --fake-clientset \
+    --clientset-name clientset \
+    --input "${input_versions}" \
```
balamurugana, line 94: Why can't we use this input argument in `openapi-gen`?

hack/update-codegen.sh
```diff
@@ -78,19 +78,22 @@ for version in "${versions[@]}"; do
...
-        --output-package "${REPOSITORY}/pkg/" \
-        --fake-clientset \
-        --clientset-name clientset \
-        --input "direct.csi.min.io/${version}" \
-        --input-base "${REPOSITORY}/pkg/apis"
-
 done
 
+# Prefix direct.csi.min.io/ to each versions.
+versions_arr=("${versions[@]/#/direct.csi.min.io/}")
```
balamurugana, line 84: reuse `arr` variable

hack/update-codegen.sh
Praveenrajmani: we need `clientset.go` to be generated with all the versions supported - `direct.csi.min.io/v1alpha1,direct.csi.min.io/v1beta1,direct.csi.min.io/v1beta2,direct.csi.min.io/v1beta3`. Currently, it supports only `direct.csi.min.io/v1beta3`.

We needed this for direct-csi integration in console.

balamurugana: The question is about `openapi-gen` execution. Can't we use it like this PR fixes?

Praveenrajmani: @balamurugana  For `openapi-gen`, we cannot have comma separated inputs for `--output-package`

wlan0 (approved):

PR #440, opened 2021-11-16
Add utility to convert DirectCSI objects to a specific version

Praveenrajmani:
Adds a utility in 'converter' pkg to convert the direct-csi object
to a specific version

This will be used for backward compatibility support in client calls


sinhaashish (approved):
Tested   & LGTM

Praveenrajmani:
Closing this PR as we don't deal with runtime objects for conversions

PR #441, opened 2021-11-16
[WIP] Add client adapter for multi-version compatibility

sinhaashish:


pkg/utils/adapter-converters.go
```diff
@@ -0,0 +1,45 @@
...
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
+
+func ErrUnsupportedConversion(from, to string) error {
+	return fmt.Errorf("Unsupported conversion %s -> %s", from, to)
+}
+
+type conversionFunc func(runtime.Object) (runtime.Object, error)
+
+func noop(from runtime.Object) (runtime.Object, error) {
```
wlan0, line 31: This should return from object. Not nil

pkg/utils/client-adapter.go
```diff
@@ -0,0 +1,206 @@
+// This file is part of MinIO Direct CSI
```
wlan0, line 1: lets rename this to drive-adapter.go

pkg/utils/client-adapter.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+const (
+	versionV1Alpha1 = "direct.csi.min.io/v1alpha1"
+	versionV1Beta1  = "direct.csi.min.io/v1beta1"
+	versionV1Beta2  = "direct.csi.min.io/v1beta2"
+	versionV1Beta3  = "direct.csi.min.io/v1beta3"
+)
+
+// var _ rest.Interface = &DirectCSDriveAdapter{}
```
wlan0, line 45: this should be uncommented

wlan0:
Let's get this to a position where it compiles successfully and continue from there. So far, the approach looks ok.

pkg/utils/client-adapter.go
```diff
@@ -0,0 +1,206 @@
...
+		converters = append(converters, drivev1alpha1ToV1beta1)
+		fallthrough
+	case directcsiv1beta1.Version:
+		converters = append(converters, drivev1beta1ToV1beta2)
+		fallthrough
+	case directcsiv1beta2.Version:
+		converters = append(converters, drivev1beta2ToV1beta3)
+	default:
+		return nil, ErrUnsupportedConversion(c.gvk.Version, directcsi.Version)
+	}
```
Praveenrajmani, line 71: We don't need `runConversionChain`, We can directly use

```
converter.ConvertDirectCSIObject(object, directcsi.Group + "/" + directcsi.Version) 
```

PR #449, opened 2021-11-19
V1.4.4 patch

wlan0:
If we upgrade from any older version to v1.4.4 or v1.4.5, and if the following holds true:
 - there are partitioned drives in use by pods
 - Uevent file does not contain partition number (PARTN)

Then, the uevent file parsing will fail. This will lead to drives showing up as `Unavailable` with a error message showing 
```
“strconv: cannot parse “”: invalid part num”
```

This failure will make us delete the existing drives because we couldn't find it in the host (because of the parsing bug). The consequence of this:
- All drives on the host will go to Terminating state.
- We won't be able to provision new volumes or kill existing pods and their PVCs.
- we won't be able to access data without manual intervention

However, there won't be any data loss.

I will follow up with a utility to recover from this issue and docs.


pkg/node/discovery/utils.go
```diff
@@ -147,17 +147,77 @@ func (d *Discovery) syncDrive(ctx context.Context, localDrive *directcsi.DirectC
...
 
 	return nil
 }
+
+func (d *Discovery) verifyDriveLost(drive *directcsi.DirectCSIDrive) (bool, error) {
+	path := sys.GetDirectCSIPath(drive.Status.FilesystemUUID)
+
+	stat, err := fs.Stat(path)
+	if err != nil {
+		if err.Unwrap() == fs.ErrNotExist {
```
balamurugana, line 179: `errors.Is(err, os.ErrNotExist)` is better way to do

Praveenrajmani:
https://github.com/Praveenrajmani/directcsi-drive-recovery-v1.4.5/blob/master/README.md

PR #450, opened 2021-11-22
Remove unused/unnecessary util functions.

balamurugana:


pkg/client/client.go
```diff
@@ -132,6 +133,15 @@ type objectResult struct {
 	err    error
 }
 
+func logYAML(obj interface{}) error {
```
wlan0, line 136: we are not using this right? can we remove this as well?

wlan0: nvm, I see this is being used

wlan0 (approved):

wlan0 (approved):

PR #451, opened 2021-11-22
Modify audit log

sinhaashish:


cmd/kubectl-direct_csi/install.go
```diff
@@ -92,22 +93,20 @@ func install(ctx context.Context, args []string) (err error) {
 	}
 	defaultAuditDir, err := utils.GetDefaultAuditDir()
 	if err != nil {
-		return fmt.Errorf("unable to get default audit directory; %w", err)
+		klog.Errorf("unable to get default audit directory : %v", err)
 	}
 	if err := os.MkdirAll(defaultAuditDir, 0700); err != nil {
-		return err
+		klog.Errorf("unable to create default audit directory : %v", err)
```
balamurugana, line 99: As you just log and ignore the error, what is this indented for?

cmd/kubectl-direct_csi/utils.go
```diff
@@ -133,22 +133,20 @@ func processFilteredDrives(
 
 	defaultAuditDir, err := utils.GetDefaultAuditDir()
```
balamurugana, line 134: Why can't we have a function to avoid this code repetition?

pkg/utils/utils.go
```diff
@@ -130,3 +132,20 @@ func GetDefaultAuditDir() (string, error) {
...
 }
+
+// Write the yaml output to a file
+func WritetoAuditFile(auditFile string) (*SafeFile, error) {
+	defaultAuditDir, err := GetDefaultAuditDir()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get default audit directory ; %w", err)
+	}
+	if err := os.MkdirAll(defaultAuditDir, 0700); err != nil {
+		return nil, fmt.Errorf("unable to create default audit directory : %v", err)
```
balamurugana, line 143: Wrap the error

pkg/utils/utils.go
```diff
@@ -130,3 +132,20 @@ func GetDefaultAuditDir() (string, error) {
...
+	}
+
+	file, err := NewSafeFile(filepath.Join(defaultAuditDir, fmt.Sprintf("%v-%v", auditFile, time.Now().UnixNano())))
+	if err != nil {
+		return nil, fmt.Errorf("unable to get audit file : %v", err)
```
balamurugana, line 148: Wrap the error and use error text as `unable to open audit file`

pkg/utils/utils.go
```diff
@@ -130,3 +132,20 @@ func GetDefaultAuditDir() (string, error) {
 	}
 	return filepath.Join(defaultDir, auditDir), nil
 }
+
+// Write the yaml output to a file
+func WritetoAuditFile(auditFile string) (*SafeFile, error) {
```
balamurugana, line 137: Name the function as `OpenAuditFile()`

cmd/kubectl-direct_csi/install.go
```diff
@@ -90,24 +88,15 @@ func install(ctx context.Context, args []string) (err error) {
...
-	}
-	if err := os.MkdirAll(defaultAuditDir, 0700); err != nil {
-		return err
-	}
 
-	file, err := utils.NewSafeFile(fmt.Sprintf("%v/%v-%v", defaultAuditDir, auditInstall, time.Now().UnixNano()))
+	file, err := utils.WritetoAuditFile(auditInstall)
 	if err != nil {
-		return fmt.Errorf("unable to get default audit directory ; %w", err)
+		return err
```
balamurugana, line 94: You don't want to ignore the error this time?

sinhaashish: logged the error

pkg/utils/utils.go
```diff
@@ -130,3 +132,19 @@ func GetDefaultAuditDir() (string, error) {
...
+func OpenAuditFile(auditFile string) (*SafeFile, error) {
+	defaultAuditDir, err := GetDefaultAuditDir()
+	if err != nil {
+		return nil, fmt.Errorf("unable to get default audit directory ; %w", err)
+	}
+	if err := os.MkdirAll(defaultAuditDir, 0700); err != nil {
+		return nil, fmt.Errorf("unable to create default audit directory : %w", err)
+	}
+
+	file, err := NewSafeFile(filepath.Join(defaultAuditDir, fmt.Sprintf("%v-%v", auditFile, time.Now().UnixNano())))
```
balamurugana, line 145: Just return without assigning to `file, err`

cmd/kubectl-direct_csi/install.go
```diff
@@ -90,24 +88,15 @@ func install(ctx context.Context, args []string) (err error) {
...
-		return fmt.Errorf("unable to get default audit directory ; %w", err)
+		klog.Errorf("error in audit logging: %w", err)
 	}
 
 	defer func() {
-		if cerr := file.Close(); err != nil {
-			klog.Errorf("unable to close file; %w", cerr)
-		} else {
-			err = cerr
+		if err := file.Close(); err != nil {
```
balamurugana, line 98: `file` can be `null` and wherever you use `file`, you have to check for `null`

balamurugana (approved):

wlan0 (approved):
This PR is OK. We should tie in context into file writing. If the context is cancelled, then file should be closed before ending the program. 

One more thing to improve is the fact that we are creating a file inside a function and expecting the caller to close the file. It is always better to utilize and cleanup resources within a function itself. This pattern we are following now is technical debt

PR #475, opened 2021-12-21
Modify version adaptability test

sinhaashish:


pkg/client/interface_test.go
```diff
@@ -33,18 +33,13 @@ import (
...
-			},
-		},
-	}
-}
+const (
+	mb20           = 20 * 1024 * 1024
+	testNodeName   = "test-node-1"
+	testDriveName  = "test-drive-1"
+	testTenantName = "tenant-1"
+	tenantLabel    = "direct.csi.min.io/tenant"
```
Praveenrajmani, line 41: we don't need this label, We don't support filtering on such labels.

For volumes, we support the following labels in filtering,

```
direct.csi.min.io/drive-path
direct.csi.min.io/node
direct.csi.min.io/pod.name
direct.csi.min.io/pod.namespace
```

Praveenrajmani (commented):
Minor nit, LGTM otherwise

wlan0 (approved):

PR #481, opened 2022-01-03
rename cmd directories and add symlinks to old paths

wlan0:


wlan0:
Merging because the failing step is codecov token is missing for directpv.

PR #488, opened 2022-01-04
add --no-header flag and hide unused persistent flags

wlan0:


functests/common.sh
```diff
@@ -194,7 +194,8 @@ function uninstall_minio() {
     "${DIRECT_CSI_CLIENT}" volumes ls
 
     while true; do
-        count=$("${DIRECT_CSI_CLIENT}" volumes ls | grep -v WARNING| wc -l)
+        count=$("${DIRECT_CSI_CLIENT}" volumes ls | grep -vc WARNING)
```
balamurugana, line 197: You should use ` | wc -l` otherwise `grep -vc` will return zero on failure.

functests/tests.sh
```diff
@@ -94,7 +94,7 @@ function do_upgrade_test() {
     # Show output for manual debugging.
     "${DIRECT_CSI_CLIENT}" volumes list --all -o wide
 
-    mapfile -t upgraded_volumes < <("${DIRECT_CSI_CLIENT}" volumes list --status published | awk '{print $1}')
+    mapfile -t upgraded_volumes < <("${DIRECT_CSI_CLIENT}" volumes list --status published | grep -v WARNING | awk '{print $1}')
```
balamurugana, line 97: Use `awk '/WARNING/ {print $1}'` instead of `grep`

functests/tests.sh
```diff
@@ -117,7 +117,7 @@ function do_upgrade_test() {
...
```
balamurugana, line 120: Same as above

functests/common.sh
wlan0: `grep | wc -l` leads to lint failure.

balamurugana: > `grep | wc -l` leads to lint failure. 

We need add to ignore list

Praveenrajmani:
Reopened as https://github.com/minio/directpv/pull/491 with the CI fixes and reviews addressed

PR #489, opened 2022-01-04
provide --build-arg=BINARY=directpv to build container

harshavardhana:


functests/common.sh
```diff
@@ -77,7 +77,7 @@ function _wait_directcsi_to_start() {
     while [[ $running_count -lt $required_count ]]; do
         echo "$ME: waiting for $(( required_count - running_count )) direct-csi pods to come up"
         sleep $(( required_count - running_count ))
-        running_count=$(kubectl get pods --field-selector=status.phase=Running --no-headers --namespace=direct-csi-min-io | wc -l)
+        running_count=$(kubectl get pods --field-selector=status.phase=Running --no-headers --namespace=directpv-min-io | wc -l)
```
Praveenrajmani, line 80: The namespace changes are not yet done. We still use `direct-csi-min-io` ns in source. Can we take care of the ns change later? It has to be done carefully as we use certs signed with the name. And also the upgrades has to be tested with these changes.

functests/tests.sh
```diff
@@ -28,22 +28,22 @@ function test_build() {
...
 }
 
 function do_upgrade_test() {
     wget --quiet --output-document="kubectl-direct_csi_$1" "https://github.com/minio/directpv/releases/download/v$1/kubectl-direct_csi_$1_linux_amd64"
     chmod a+x "kubectl-direct_csi_$1"
 
-    # unmount all direct-csi mounts of previous installation if any.
-    mount | awk '/direct-csi/ {print $3}' | xargs sudo umount -fl
+    # unmount all directpv mounts of previous installation if any.
+    mount | awk '/directpv/ {print $3}' | xargs sudo umount -fl
```
Praveenrajmani, line 40: older mounts from versions < 1.5 would still have `/direct-csi/` in the mountpaths

functests/tests.sh
```diff
@@ -28,22 +28,22 @@ function test_build() {
...
 
     DIRECT_CSI_CLIENT="./kubectl-direct_csi_$1"
     DIRECT_CSI_VERSION="v$1"
-    image="direct-csi:${DIRECT_CSI_VERSION}"
+    image="directpv:${DIRECT_CSI_VERSION}"
     if [ -n "${RHEL7_TEST}" ]; then
-        image="direct-csi:${DIRECT_CSI_VERSION}-rhel7"
+        image="directpv:${DIRECT_CSI_VERSION}-rhel7"
```
Praveenrajmani, line 46: the new tag name `directpv:<version>` doesn't apply for older versions. (v1.3.6, v1.4.6 - used for upgrade tests)

functests/common.sh
wlan0: Yeah, this shouldn't be changed

PR #502, opened 2022-01-11
Run dynamic drive discovery service as separate container.

balamurugana:

### DIFF
--- a/pkg/installer/daemonset.go
+++ b/pkg/installer/daemonset.go
@@ -225,6 +225,40 @@ func createDaemonSet(ctx context.Context, c *Config) error {
 	if c.DynamicDriveDiscovery {
 		podSpec.HostNetwork = true
 		podSpec.DNSPolicy = corev1.DNSClusterFirstWithHostNet
+		args := []string{
+			fmt.Sprintf("--identity=%s", c.daemonsetName()),
+			fmt.Sprintf("-v=%d", logLevel),
+			fmt.Sprintf("--endpoint=$(%s)", endpointEnvVarCSI),
+			fmt.Sprintf("--node-id=$(%s)", kubeNodeNameEnvVar),
+			"--dynamic-drive-handler",
+		}
+		if c.LoopbackMode {
+			args = append(args, "--loopback-only")
+		}
+		podSpec.Containers = append(podSpec.Containers, corev1.Container{
+			Name:            directPVDriveDiscoveryContainerName,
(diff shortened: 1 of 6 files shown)


pkg/node/uevent.go
```diff
@@ -418,3 +418,28 @@ func (handler *ueventHandler) processLoop(ctx context.Context) {
 		handler.processEvent(ctx, device, event["ACTION"])
 	}
 }
+
+func StartDynamicDriveDiscovery(ctx context.Context, identity, nodeID, rack, zone, region string, loopbackOnly bool) error {
+	handler := &ueventHandler{
+		nodeID: nodeID,
+		topology: map[string]string{
```
wlan0, line 425: do we need topology?

pkg/installer/daemonset.go
```diff
@@ -146,7 +146,7 @@ func createDaemonSet(ctx context.Context, c *Config) error {
 						args = append(args, "--loopback-only")
 					}
 					if c.DynamicDriveDiscovery {
-						args = append(args, "--dynamic-drive-discovery")
+						args = append(args, "--enable-dynamic-discovery")
```
Praveenrajmani, line 149: should we support backward compatility here? renaming this flag will break backward compatibility. 

If v2.0.1 client is used to install, `--dynamic-drive-discovery` will be set on the binary.

pkg/node/node.go
```diff
@@ -93,32 +93,6 @@ func NewNodeServer(ctx context.Context, identity, nodeID, rack, zone, region str
...
-				return nil, err
-			}
-		}
-
-		klog.V(3).Info("Doing initial drive sync up")
-		handler.syncDrives(ctx)
-
-		// Start background tasks
-		go handler.processLoop(ctx)
-	}
```
Praveenrajmani, line 120: if we decide to support backward compatibility, we need to keep this code. we need to make this code segment run, if `--dynamic-drive-discovery` flag is set.

pkg/node/uevent.go
Praveenrajmani: yes, we need topology as uevent handler creates new drive objects if any drives are added

pkg/installer/daemonset.go
wlan0: yeah, lets keep the existing flag

pkg/node/node.go
wlan0: It is going to run as a separate container. This is not needed.

wlan0:
LGTM, only one minor comment about flag. please fix, and it can be merged

wlan0 (approved):

PR #506, opened 2022-01-12
change to directpv in functional tests

sinhaashish:


Praveenrajmani (commented):
we can modify and run the add-license-header.sh script instead of manually updating the files

build.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# This file is part of MinIO Direct CSI
+# This file is part of MinIO Direct PV
```
Praveenrajmani, line 3:
```suggestion
# This file is part of MinIO DirectPV
```

PR #508, opened 2022-01-13
Use 'directpv-min-io' namespace for installations and few renames

Praveenrajmani:


pkg/installer/config.go
```diff
@@ -49,10 +52,10 @@ func defaultIfZeroString(left, right string) string {
 type Config struct {
 	Identity string
 
-	// DirectCSIContainerImage properties
-	DirectCSIContainerImage    string
-	DirectCSIContainerOrg      string
-	DirectCSIContainerRegistry string
+	// DirectPVContainerImage properties
```
wlan0, line 55: Is this comment needed?

wlan0 (approved):

pkg/installer/config.go
Praveenrajmani: keeping the comment as it was there already @wlan0

kaplanlior:
Thanks for the patch. I hope it would be accepted soon.

Some parts of the code still mentions direct-csi, search _directcsi and DIRECT_CSI for those.

Notice the function tests (installation / upgrade) still look for direct-csi and binaries with that name.

kaplanlior:
Another comment - I'm not sure where in the code it's set, but on the nodes, the mount points are using direct-csi as their path
e.g. 
```
/var/lib/direct-csi/mnt/3c32999c-5f37-4256-b1ed-5e247d05ecf6
```

kaplanlior:
Checked version 3.0.0 and it still uses direct-csi-min-io for name space and other resources.

Praveenrajmani:
closing this as the PR has become stale, will open a fresh PR later

PR #509, opened 2022-01-13
Updated License header with DirectPV

sinhaashish:


Praveenrajmani (dismissed):

build.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-# This file is part of MinIO Direct CSI
-# Copyright (c) 2021 MinIO, Inc.
+# This file is part of MinIO DirectPV
+# Copyright (c) 2022 MinIO, Inc.
```
balamurugana, line 4: You should retain `2021` and previous years copyright. It should be like `2021, 2022`. It is applicable everywhere except files those are all appearing from `2022`.

cmd/kubectl-directpv/drives.go
```diff
@@ -1,20 +1,18 @@
...
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU Affero General Public License for more details.
- *
- * You should have received a copy of the GNU Affero General Public License, version 3,
- * along with this program.  If not, see <http://www.gnu.org/licenses/>
- *
- */
+// This file is part of MinIO DirectPV
```
balamurugana, line 1: Just minimize the diff by changing respective lines. Do wherever applicable.

pkg/apis/direct.csi.min.io/v1alpha1/deepcopy_generated.go
```diff
@@ -1,19 +1,19 @@
...
-//
+// This file is part of MinIO DirectPV
+// Copyright (c) 2022 MinIO, Inc.
+
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
 // the Free Software Foundation, either version 3 of the License, or
 // (at your option) any later version.
-//
+
```
balamurugana, line 11: Are you regenerated the code with new header having updated name and year or editing manually?

balamurugana (changes requested):

cmd/kubectl-directpv/drives.go
harshavardhana: This is on purpose Go's recommended way is to use `//` instead of `/*` for license header.

Almost all linters complain about this.

balamurugana: Can you point me about the recommendation to know why `golangci-lint` workflow doesn't complain that?

harshavardhana: golangci-lint requires you to enable other linters that are not yet enabled @balamurugana Linters are a non exhaustive list 

That main reason is using `/*` conflicts with Go's build tags and their actual positions in the file.

This is why MinIO, Console and mc are all now using `//` instead of `/*` for license headers.

DirectPV should follow this as well.

balamurugana: This should be done as separate PR along with enabling such linters in golangci-lint workflow. It is a bad idea reviewers to do linter work.

harshavardhana: No this PR is fine as is. We should take this PR as is. It fixes something valid.

Your argument is not valid here, I don't see why it should be done in separate PRs - it's not like it's fixing a bug.

build.sh
sinhaashish: checked all PR merged in 2022. Didnt find any new files added.  modified to `2021, 2022` in all the files.

cmd/kubectl-directpv/drives.go
sinhaashish: @balamurugana  I ran  `find . -name "*.go" -print0 | xargs --null addlicense -f "${SCRIPT_DIR}/boilerplate.go.txt"` . It adds comment as `//` . IMO we can take this PR as `/*` -> `//` just increases the diff nothing else , In many go files i saw `//` being used .

build.sh
balamurugana: Files added prior to 2022 should have retain previous years copyright and files added in 2022 shouldn't for obvious reason.

pkg/apis/direct.csi.min.io/v1beta1/conversion_generated.go
```diff
@@ -14,6 +14,7 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU Affero General Public License for more details.
 //
+
```
balamurugana, line 17: Remove this newline

cmd/kubectl-directpv/drives.go
balamurugana: @harshavardhana We should add specified change with enabled linter. If you want it in this PR, I am fine. This is only the way, multiple commented license header will be reported and fixed in upcoming PRs

.golangci.yml
```diff
@@ -10,6 +10,23 @@ linters-settings:
   misspell:
     locale: US
 
+linters:
+  disable-all: true
```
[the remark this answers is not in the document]
sinhaashish: This is what is done in `minio` and `mc`. I didn't find any specific go linter which  present in mc or minio which handles multilined comment .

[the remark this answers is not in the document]
balamurugana: You would need to check what linter is doing the job. If no error means it is allowed to use single and multi-line comments as license header.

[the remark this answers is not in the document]
sinhaashish: @harshavardhana  Can you please tell which specific linter blocks multi line comments?

pkg/sys/smart/smart_test.go
```diff
@@ -16,6 +14,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+//go:build !skip
```
balamurugana, line 17: Move all build constraints to beginning of the file.

.golangci.yml
```diff
@@ -10,6 +10,23 @@ linters-settings:
...
+    - govet
+    - ineffassign
+    - gosimple
+    - deadcode
+    - structcheck
+    - gomodguard
+    - gofmt
+    - unused
+    - structcheck
+    - varcheck
```
balamurugana, line 28: Aren't these checkers enabled by default?

pkg/sys/smart/smart_test.go
sinhaashish: build constraints at the top leads to goheader lint failure.

.golangci.yml
balamurugana: It seems many usable linters are disabled by default. We will enable and fix respectively in different PR.

pkg/sys/smart/smart_test.go
balamurugana: Weird I have never seen this style anywhere. You would need to fix it accordingly.

.golangci.yml
sinhaashish: The default linters are

pkg/sys/smart/smart_test.go
sinhaashish: done

balamurugana (approved):

pkg/sys/smart/smart_test.go
harshavardhana: build constraints are not the topmost header - https://github.com/minio/kes/blob/master/internal/fips/fips.go#L5 check how to add them.

It was an older Go code that had this and some packages in runtime library are still in the old format. Backward compatibility is still maintained.

But they moved to build tags right before the `package <name>`

harshavardhana (approved):

PR #511, opened 2022-01-17
add index.html file

sinhaashish:
The first line of read me is not the same as repo name , so the hypelink is coming shown below 
Screenshot from 2022-01-17 15-30-44

The solution is to have an idex.html created out of readme.md. Refer https://stackoverflow.com/questions/47104807/github-pages-homepage-remove-link-at-the-top

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<!--- Please update index.html if readme.md is updated --->
+
 DirectPV
 ----------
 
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

sinhaashish:
Note :
Update the `index.html` file every time you update the `readme.md`

PR #522, opened 2022-01-31
Add support for kubernetes v1.23

Praveenrajmani:
Fixes #507

PR #523, opened 2022-02-01
Mount drive only if CRD drive object succeeded in sync drive

balamurugana:


Praveenrajmani:
```
 error: cannot update repo 'base-8': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Status code: 404 for https://mirrors.edge.kernel.org/centos/8/BaseOS/x86_64/os/repodata/repomd.xml (IP: 147.75.197.195)
 ```
 looks like the cdn is down. also see this problem in https://github.com/minio/directpv/pull/522

PR #524, opened 2022-02-01
Sync drives before starting the node server

Praveenrajmani:
- currently, the node server will start parallelly with dynamic discovery container.
- this will cause some inconsistencies in creating, staging and publishing volumes
- initial drive state sync is required as the node server is depended on the drive states

### DIFF
--- a/pkg/node/drive_sync.go
+++ b/pkg/node/drive_sync.go
@@ -1,0 +1,159 @@
+// This file is part of MinIO DirectPV
+// Copyright (c) 2021, 2022 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package node
+
+import (
+	"context"
+	"strings"
+
+	"github.com/google/uuid"
+	directcsi "github.com/minio/directpv/pkg/apis/direct.csi.min.io/v1beta3"
+	"github.com/minio/directpv/pkg/client"
+	"github.com/minio/directpv/pkg/sys"
+	"github.com/minio/directpv/pkg/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+)
+
+func syncDrives(ctx context.Context, nodeID string, loopbackOnly bool, topology map[string]string) error {
+	devices, err := sys.ProbeDevices()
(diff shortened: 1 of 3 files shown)


cmd/directpv/run.go
```diff
@@ -185,6 +185,10 @@ func run(ctx context.Context, args []string) error {
 				return fmt.Errorf("error while initializing drive discovery: %v", err)
 			}
 			klog.V(3).Infof("Drive discovery finished")
+		} else {
+			if err := node.SyncDrives(ctx, identity, nodeID, rack, zone, region, loopbackOnly); err != nil {
```
balamurugana, line 189: This doesn't make sense when dynamic drive discovery runs as separate container. The container does sync first and waits/processes uevents.

Praveenrajmani: when dynamic discovery runs as a separate container, it runs alongside with the node server container. So, node server will be running parallel with dynamic-discovery container. 

This may lead to staging and publish volume calls to happen event before the first sync of the drives.

@balamurugana

pkg/node/node.go
```diff
@@ -94,6 +94,12 @@ func NewNodeServer(ctx context.Context, identity, nodeID, rack, zone, region str
 		setQuota:        xfs.SetQuota,
 	}
 
+	if dynamicDriveDiscovery {
+		if err := syncDrives(ctx, identity, nodeID, rack, zone, region, loopbackOnly); err != nil {
+			return &NodeServer{}, err
```
wlan0, line 99: Log the error, and return nil, err

pkg/node/uevent.go
```diff
@@ -439,6 +439,23 @@ func StartDynamicDriveHandler(ctx context.Context, identity, nodeID, rack, zone,
 		dynamicDriveDiscovery: true,
 		loopbackOnly:          loopbackOnly,
 	}
+
+	return handler.processLoop(ctx)
+}
+
+func syncDrives(ctx context.Context, identity, nodeID, rack, zone, region string, loopbackOnly bool) error {
```
wlan0, line 446: sync drives should not need a handler. We are not hadling uevents for initial sync

wlan0 (changes requested):
Add test cases and refactor sync drives to run independent of uevent handler

balamurugana (approved):

wlan0:
This PR needs to be simplified further. It is not easy to tell what `syncDrive` vs. what `syncDrives` does. The names are not inutitive.

pkg/node/drive_sync.go
```diff
@@ -0,0 +1,159 @@
...
+		nil,
+		nil,
+		client.MaxThreadCount,
+	)
+	if err != nil {
+		klog.Error(err)
+		return err
+	}
+
+	for result := range resultCh {
```
wlan0, line 52: This is wrong. I wanted to get all the drives first before trying to match. By doing that, we ensure we can do elimination logic for matching.

pkg/node/drive_sync.go
```diff
@@ -0,0 +1,159 @@
...
+		if result.Err != nil {
+			klog.Error(result.Err)
+			return result.Err
+		}
+
+		if matched, updated := updateDrive(ctx, &result.Drive, devices); matched {
```
wlan0, line 58: Do not return matched, updated from updateDrive call. Break down into three separate steps.

wlan0 (changes requested):
Too many bad design choices. I'm closing this. We have to rewrite this module.

PR #525, opened 2022-02-01
Add filesystem UUID check before staging the volume

Praveenrajmani:


pkg/node/utils.go
```diff
@@ -17,16 +17,43 @@
...
 	"github.com/minio/directpv/pkg/mount"
 	"github.com/minio/directpv/pkg/sys"
+	"github.com/minio/directpv/pkg/utils"
+	"k8s.io/klog/v2"
 )
 
-func checkDrive(drive *directcsi.DirectCSIDrive, volumeID string, probeMounts func() (map[string][]mount.Info, error)) error {
+func (n *NodeServer) checkDrive(ctx context.Context, drive *directcsi.DirectCSIDrive, volumeID string) error {
+
+	probeXFSUUID := func(majMin string) (string, error) {
```
wlan0, line 34: why is a closure needed here?

wlan0:
Can you add tests for cases where the right drive is mounted at the mountpoint, the wrong drive is mounted at the mountpoint and no drive is mounted at that mountpoint

wlan0 (changes requested):
Remove unnecessary closures and add testcases

PR #527, opened 2022-02-02
Update the mountpoint after mounting the drive on node restart

Praveenrajmani:
Currently, as the `drive.Status.Mountpoint` is left empty after mounting, the volumes will be bind mounted to the `/` dir (/pvc-XXX)

This patch will fix the volume mounts on node restarts.

### DIFF
--- a/pkg/node/uevent.go
+++ b/pkg/node/uevent.go
@@ -39,29 +39,40 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func mountDrive(ctx context.Context, drive *directcsi.DirectCSIDrive) {
-	target := filepath.Join(sys.MountRoot, drive.Status.FilesystemUUID)
-	var flags []string
-	if drive.Spec.RequestedFormat != nil {
-		flags = drive.Spec.RequestedFormat.MountOptions
-	}
-	err := mount.MountXFSDevice(drive.Status.Path, target, flags)
-	if err == nil {
-		return
-	}
-
-	klog.ErrorS(err, "unable to mount drive", "Status.Path", drive.Status.Path, "Target", target, "Flags", flags)
-	utils.UpdateCondition(
-		drive.Status.Conditions,
-		string(directcsi.DirectCSIDriveConditionInitialized),
-		utils.BoolToCondition(false),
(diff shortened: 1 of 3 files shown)


pkg/node/uevent.go
```diff
@@ -39,29 +39,42 @@ import (
...
-		string(directcsi.DirectCSIDriveReasonInitialized),
-		err.Error(),
-	)
+func mountDrive(ctx context.Context, drive *directcsi.DirectCSIDrive, mountFn func(device, dest string, mountflags []string) error) {
 	driveInterface := client.GetLatestDirectCSIDriveInterface()
-	err = retry.RetryOnConflict(
+	err := retry.RetryOnConflict(
 		retry.DefaultRetry,
 		func() (err error) {
+			drive, gErr := driveInterface.Get(
```
balamurugana, line 47: why can't we use `err` itself here?

pkg/node/uevent.go
```diff
@@ -39,29 +39,42 @@ import (
...
+			}
+			target := filepath.Join(sys.MountRoot, drive.Status.FilesystemUUID)
+			var flags []string
+			if drive.Spec.RequestedFormat != nil {
+				flags = drive.Spec.RequestedFormat.MountOptions
+			}
+			message := ""
+			err = mountFn(drive.Status.Path, target, flags)
+			if err == nil {
+				// successfully mounted (OR) already mounted
```
balamurugana, line 61: This comment is obvious and just remove it

pkg/node/uevent_helpers.go
```diff
@@ -193,7 +193,7 @@ func updateDriveProperties(drive *directcsi.DirectCSIDrive, device *sys.Device)
 		updated = true
 	}
 
-	if drive.Status.Mountpoint != device.FirstMountPoint {
+	if device.FirstMountPoint != "" && drive.Status.Mountpoint != device.FirstMountPoint {
```
balamurugana, line 196: Time being this check is OK. However we should update if `FSUUID` is updated above.

balamurugana (approved):

PR #540, opened 2022-03-10
Fix checkXFS check to handle "operation not supported" mount errors as well

Praveenrajmani:
check mounting with reflink enable during formatting. If the mounting fails, try with reflink disabled and conclude based on the results


wlan0 (approved):

wlan0:
LGTM

sinhaashish (commented):
code looks got, not able to test against SUSE 12.5

PR #545, opened 2022-03-15
Ignore device probe errors while initializing and continue discovering other devices

Praveenrajmani:
currently, we error out even if one device probe fails. This is a fix to ignore device probe errors and continue probing the other devices.


harshavardhana:
This is not a correct fix @Praveenrajmani you need to ignore device based on `/hidden` value 

https://github.com/util-linux/util-linux/commit/c8487d854ba5cf5bfcae78d8e5af5587e7622351 read how to do this here.

harshavardhana:
This PR https://github.com/minio/directpv/pull/546 should replace this.

harshavardhana (dismissed):

Praveenrajmani:
We need this PR to ignore/skip any I/O errors while probing devices

harshavardhana (approved):

PR #546, opened 2022-03-15
read only relevant device properties when 'hidden'

harshavardhana:


Praveenrajmani (dismissed):
LGTM

sinhaashish (dismissed):

Praveenrajmani:
few things to look at 
- we need to make hidden devices as `Unavailable` - https://github.com/minio/directpv/blob/master/pkg/client/drive.go#L50
- we need to see if fsprobe should also be skipped for such hidden devices

can do these things in a separate PR as well

Praveenrajmani (approved):

sinhaashish (approved):

PR #581, opened 2022-05-11
Sort volume list by PODNAMESPACE

sinhaashish:
The table has the sort functionality defined here "github.com/jedib0t/go-pretty/v6/table".

Refer https://github.com/jedib0t/go-pretty/blob/9c7a814efc0e7394783a7d7c1682c3ed068846f2/table/sort_test.go#L31

### DIFF
--- a/cmd/kubectl-directpv/volumes_list.go
+++ b/cmd/kubectl-directpv/volumes_list.go
@@ -163,6 +163,13 @@ func listVolumes(ctx context.Context, args []string) error {
 		}
 		t.AppendRow(row)
 	}
+	t.SortBy(
+		[]table.SortBy{
+			{
+				Name: "PODNAMESPACE",
+				Mode: table.Asc,
+			},
+		})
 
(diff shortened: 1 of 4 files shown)


Praveenrajmani:
can we sort the volumes by pod-name instead of pod-namespace @sinhaashish ? namespaces will not be contiguos mostly, whereas the podnames are sequenced and contiguous and makes a comparative difference in listing

sinhaashish:
> can we sort the volumes by pod-name instead of pod-namespace @sinhaashish ? namespaces will not be contiguos mostly, whereas the podnames are sequenced and contiguous and makes a comparative difference in listing

We had discussed that sorting by namespaces will lead to a group by namespace where in we can check the different pod names. Like if we have namespace as production , UAT, development. the listing can be done which shows all the pods in production , UAT and development name spaces respectively. This leads to better readibility.

Praveenrajmani (approved):

PR #585, opened 2022-05-17
Honour SYSTEMD_READY udev data value

Praveenrajmani:
If this value is set to "0", then device seems to be unplugged or not initialized yet.
This uninitialized state will initially show up after which a "changed" event will be generated after the device is fully setup.

reference: https://www.freedesktop.org/software/systemd/man/systemd.device.html

```
SYSTEMD_READY=

    If set to 0, systemd will consider this device unplugged even if it shows up in the udev tree. If this property is unset or set to 1, the device will be considered plugged if it is visible in the udev tree.

    This option is useful for devices that initially show up in an uninitialized state in the tree, and for which a "changed" event is generated the moment they are fully set up. Note that SYSTEMD_WANTS= (see above) is not acted on as long as SYSTEMD_READY=0 is set for a device.
```

### DIFF
--- a/functests/tests.sh
+++ b/functests/tests.sh
@@ -147,6 +147,8 @@ echo "$ME: Setup environment"
 setup_lvm
 setup_luks
 
+sleep 30
+
 echo "$ME: Run build test"
 test_build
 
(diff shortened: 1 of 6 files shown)


Praveenrajmani:
this value cannot be used for LUKS devices

PR #606, opened 2022-06-23
Add a flag to disable udev listener

Praveenrajmani:
setting `--disable-udev-listener` during install will disable udev monitoring. This will skip
listening for the udev Add, Change and Remove events and relies on internal sync mechanism in-place
to sync the drive states.

Steps to test :
- run `./build.sh`
- build and push the image
- ./kubectl-directpv install --disable-udev-listener --org <org> --image <image>
- basic tests like formatting, releasing drives (This should take minimum 30secs to reflect)

### DIFF
--- a/cmd/directpv/cmd.go
+++ b/cmd/directpv/cmd.go
@@ -51,6 +51,7 @@ var (
 	dynamicDriveDiscovery = false
 	dynamicDriveHandler   = false
 	metricsPort           = 10443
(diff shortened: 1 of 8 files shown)


sinhaashish (dismissed):
Just to make a point , that we should provide the details as to why this change is needed. Other wise LGTM

balamurugana (approved):

PR #609, opened 2022-07-04
Fix swap partition discoveries and some minor bug-fixes

Praveenrajmani:
This PR fixes the following minor bugs

- Currently (without this PR), with swap partitions in the setup, you will see the following logs in directpv-drive-discovery container
```
 [dm-1] filesystem mismatch: linux-swap -> swap
 ```
This is because of the mismatch in udev fs probe ("swap") and fs probe by opening the device ("linux-swap")

- LVM2_member drives should be treated as "unavailable" drives
- Display appropriate errors on the disks when a inuse|ready drive gets umounted or when these drives have no fs.

Steps to test :-

- To enable swapons, you need to execute the folllowing steps

```bash
sudo mkswap <dev> #sudo mkswap /dev/dm-0
sudo swapon <dev> #sudo swapon /dev/dm-0
```
to disable, 

```bash
sudo swapoff -v <dev> #sudo swapoff -v /dev/dm-0
sudo mkfs.xfs -f <dev> #sudo mkfs.xfs -f /dev/dm-0
```
Swap drives should show up as unavailable

- Umount in-use/ready drives, the `kubectl directpv drives list` should diplay "the directpv mount is not found" instead of mismatch error text


pkg/uevent/match_test.go
```diff
@@ -329,8 +332,11 @@ func TestFileSystemTypeMatcher(t *testing.T) {
 	case1Drive := directcsi.DirectCSIDrive{Status: directcsi.DirectCSIDriveStatus{Filesystem: ""}}
 	case2Drive := directcsi.DirectCSIDrive{Status: directcsi.DirectCSIDriveStatus{Filesystem: "xfs"}}
 	case3Drive := directcsi.DirectCSIDrive{Status: directcsi.DirectCSIDriveStatus{Filesystem: "ext64"}}
```
harshavardhana, line 334:
```suggestion
	case3Drive := directcsi.DirectCSIDrive{Status: directcsi.DirectCSIDriveStatus{Filesystem: "ext64"}}
```
what is this? - is this a typo?

harshavardhana (changes requested):

pkg/uevent/match_test.go
Praveenrajmani: yes @harshavardhana will fix

PR #610, opened 2022-07-11
Move to state based matching algorithm

Praveenrajmani:
```
follows group based match, managed (ready|inuse) and non-managed (available|unavailable|released),
this isolates them and do-not allow matching collision between two groups

Managed directcsidrives (InUse and Ready) :-
=========================================
(*) Compares the FS attributes (fstype and fsuuid) in the devices and the directcsidrive.
(*) Matched directcsidrive will be updated.
(*) Unmatched directcsidrive will be tagged as "lost".
(*) Too many device matches will be logged as an error.
(*) unmatched devices will be used for the next directcsidrive match in the iteration.

Non-Managed directcsidrives (Available, Unavailable, Released) :-
==============================================================
(*) Tries to match the directcsidrive by the following attributes if available on the directcsidrive
    (1)  PartitionNumber
    (2)  WWID
    (3)  Serial
    (4)  SerialLong
    (5)  DMUUID
    (6)  MDUUID
    (7)  ModelNumber
    (8)  Vendor
    (9)  PartitionUUID
    (10) UeventFSUUID
    (11) FilesystemUUID (only matcher for v1.4.6 drives)
(*) Matched directcsidrive will be updated.
(*) Unmatched directcsidrive will be considered as un-identified directcsidrive. These
    directcsidrive could be non-upgraded or empty drives (virtual) with no persistent attributes in them
(*) Too many device matches will be logged as an error.
(*) Unmatched devices will be used for the next directcsidrive match in the iteration.

Un-identified directcsidrives (unmatched non-managed directcsidrives) :-
=====================================================================
(*) Tries to match the directcsidrive by the following attributes on the directcsidrive
     (1) drive path
     (2) major and minor number
     (3) PCIPath (if present)
(*) Matched directcsidrive will be updated.
(*) Unmatched directcsidrive will be deleted
(*) Too many device matches will be logged as an error
(*) Unmatched devices will be created
```


Praveenrajmani:
I have tested the following scenarios, please let me know if i missed anything

- Upgrade test from 1.4.6 
- Upgrade test from 1.4.3 -> 1.4.6 and then to latest
- Upgrade test from 2.0.9
- Upgrade test from 3.0.2
- Fresh installation test
(all of the above tests are done on both minikube and aws cluster)
- Attach/Detach drive test on minikube
- Attach/Detach drive test on centos7 VM
- Detaching inuse drives on centos7 VM
- tested node reboots (AWS cluster)

balamurugana (approved):

Praveenrajmani:
PTAL @sinhaashish

pkg/uevent/sync.go
```diff
@@ -0,0 +1,237 @@
...
+
+	return l.syncDevices(ctx, devices)
+}
+
+// Tries to match the directcsidrives by the discovered and probed devices from host
+//
+// follows group based match, managed (ready|inuse) and non-managed (available|unavailable|released),
+// this isolates them and do-not allow matching collision between two groups
+//
+// Managed directcsidrives (InUse and Ready) :-
```
sinhaashish, line 124: IMO this logic of matching only fs attribute  for maned drive should also be a part of architecture.mc docs

sinhaashish (approved):
Just a single comment otherwise LGTM

pkg/uevent/sync.go
Praveenrajmani: sure, will add this in the arch docs @sinhaashish

PR #618, opened 2022-08-02
cleanup code with 'gofumpt'

harshavardhana:
for platform independent behavior use
path.Join() instead of filepath.Join()

Bonus: move to go1.18.x for all builds
we have been making releases for DirectPV
with go1.18.x for last 3 releases.


Praveenrajmani (approved):

balamurugana (approved):
LGTM

PR #620, opened 2022-08-09
Additional flags to `kubectl directpv volumes list` to list "lost" volumes and to show PVC names of the volumes

Praveenrajmani:
* `--lost` will list only the lost volumes, ie, volumes from the detached drives
* `--pvc` will display PVC names of the volumes in the list

This PR also adds documentation for handling "InUse" drive replacements


sinhaashish (approved):
LGTM

Praveenrajmani:
merging this PR as we need the documentation work to be shared, will address the review comments in a separate PR if there are any

PR #643, opened 2022-09-17
fix kubectl directpv plugin command line arguments

balamurugana:


cmd/kubectl-directpv/info.go
```diff
@@ -185,7 +180,7 @@ func getInfo(ctx context.Context, args []string, quiet bool) error {
 
 		if driveCount == 0 {
 			t.AppendRow([]interface{}{
-				fmt.Sprintf("%s %s", color.RedString(dot), n),
+				fmt.Sprintf("%s %s", color.GreenString(dot), n),
```
Praveenrajmani, line 183: Can we keep the `red` string - It is to indicate that no drives are added yet to DirectPV

Praveenrajmani (approved):
a minor nit

cmd/kubectl-directpv/info.go
balamurugana: `Red` seems error. I will change it to `Yellow`

PR #645, opened 2022-09-20
Add volume commands

balamurugana:


cmd/kubectl-directpv/volumes.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func getNodeSelectors() ([]types.LabelValue, error) {
+	for i := range nodeArgs {
+		if utils.TrimDevPrefix(nodeArgs[i]) == "" {
+			return nil, fmt.Errorf("empty node name %v", nodeArgs[i])
+		}
+	}
+	return getSelectorValues(nodeArgs)
+}
```
Praveenrajmani, line 98: can these functions  move to common utils.go file as it can be used for drive listing as well?

Praveenrajmani (approved):

PR #660, opened 2022-10-21
Probe non-device name mounts

balamurugana:
Devices may be mounted using different names e.g. /dev/root. This PR fixes to probe such mounts.

### DIFF
--- a/pkg/drive/event.go
+++ b/pkg/drive/event.go
@@ -29,6 +29,7 @@ import (
 	"github.com/minio/directpv/pkg/listener"
 	"github.com/minio/directpv/pkg/sys"
 	"github.com/minio/directpv/pkg/types"
+	"github.com/minio/directpv/pkg/utils"
 	"github.com/minio/directpv/pkg/xfs"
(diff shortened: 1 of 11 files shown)


pkg/drive/event.go
```diff
@@ -174,8 +178,8 @@ func (handler *driveEventHandler) unmountDrive(ctx context.Context, drive *types
 	if len(devices) > 1 {
 		return fmt.Errorf("multiple devices %v are mounted for FSUUID %v", devices, drive.Status.FSUUID)
 	}
-	mountpoints := deviceMap[devices[0]]
-	for _, mountPoint := range mountpoints {
+	mountpoints := deviceMap[devices.ToSlice()[0]]
```
Praveenrajmani, line 181:
```go
mountpoints := deviceMap[devices[0]]
```

would also work here right?

Praveenrajmani (approved):

pkg/drive/event.go
balamurugana: No. Because it returns `map[string]struct{}`

PR #664, opened 2022-10-26
Rename api-server to admin-server

balamurugana:

### DIFF
--- a/pkg/installer/api-server-deployment.go
+++ b/pkg/installer/api-server-deployment.go
@@ -71,10 +71,10 @@ func createAPIServerDeployment(ctx context.Context, c *Config) error {
 		ImagePullSecrets: c.getImagePullSecrets(),
 		Containers: []corev1.Container{
 			{
-				Name:  apiServerContainerName,
+				Name:  consts.AdminServerName,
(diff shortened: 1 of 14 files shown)


pkg/installer/api-server-deployment.go
```diff
@@ -71,10 +71,10 @@ func createAPIServerDeployment(ctx context.Context, c *Config) error {
 		ImagePullSecrets: c.getImagePullSecrets(),
 		Containers: []corev1.Container{
 			{
-				Name:  apiServerContainerName,
+				Name:  adminServerName,
```
Praveenrajmani, line 74: can we also rename the corresponding function and filename?

cmd/directpv/api-server.go
```diff
@@ -24,9 +24,9 @@ import (
 
 var apiPort = consts.APIPort
 
-var apiServer = &cobra.Command{
-	Use:           "api-server",
-	Short:         "Start API server of " + consts.AppPrettyName + ".",
+var adminServerCmd = &cobra.Command{
```
Praveenrajmani, line 27: can we rename the file as well?

Praveenrajmani (approved):

PR #683, opened 2022-12-06
Rename namespace to directpv

balamurugana:


pkg/consts/consts.go
```diff
@@ -37,9 +37,6 @@ const (
 	// Identity denotes identity value.
 	Identity = AppName + "-min-io"
 
-	// Namespace denotes the namespace where the app is installed
-	Namespace = Identity
```
Praveenrajmani, line 41: Can we keep `Namespace = AppName` so that it will be easier in the future for modifications.

balamurugana: That's how it is https://github.com/balamurugana/directpv/blob/Rename-namespace-to-directpv/pkg/installer/consts.go#L25

harshavardhana (approved):

PR #693, opened 2022-12-30
[WIP] Add progress UI for installation

Praveenrajmani:


harshavardhana:
Is there an animation in asciinema that can be shared here @Praveenrajmani ?

Praveenrajmani:
https://asciinema.org/a/oAijSHzJpwGDOexcZvbrBDX7x

^^ @harshavardhana

pkg/installer/installer.go
```diff
@@ -113,137 +119,93 @@ func Install(ctx context.Context, args *Args) error {
...
-		}
-	}
-
-	execute := func(name string, fn func(context.Context, *Args) error) error {
-		print(name + " ")
-
-		if err := fn(ctx, args); err != nil {
-			print(cross)
-			return err
+	execute := func(fn func(context.Context, *Args) error, totalSteps int) (err error) {
```
balamurugana, line 122: We don't need this closure. Each component should accept `Args` , send start/progress/end messages. Because each component knows how many steps will be performed. They are fixed value.

`install` command knows how many components will be installed based on legacy flag. The progress at `install` command should initiate the value accordingly instead of having `TotalTasks` const.

balamurugana (approved):

PR #697, opened 2023-01-10
Improve the command helpers and minor bug fixes

Praveenrajmani:


cmd/kubectl-directpv/clean.go
```diff
@@ -35,65 +35,65 @@ import (
...
-	Use:           "release [VOLUME ...]",
+var cleanCmd = &cobra.Command{
+	Use:           "clean [VOLUME ...]",
 	Short:         "Cleanup stale volumes",
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	Example: strings.ReplaceAll(
-		`# Release all stale volumes
-$ kubectl {PLUGIN_NAME} release --all
+		`1) Cleanup all stale volumes
```
balamurugana, line 44: Use `mc` style like `1. Cleanup ...`

cmd/kubectl-directpv/init_config_v1.go
```diff
@@ -18,6 +18,10 @@ package main
 
 import directpvtypes "github.com/minio/directpv/pkg/apis/directpv.min.io/types"
 
+const (
+	selectValueYes = "yes"
```
balamurugana, line 22: if this contant is used at only one place, there is no need to declare

pkg/installer/args.go
```diff
@@ -52,6 +52,7 @@ type Args struct {
 	csiProvisionerImage      string
 	nodeDriverRegistrarImage string
 	livenessProbeImage       string
+	Printer                  func(interface{})
```
balamurugana, line 55: There is no need to output other than yaml format.

cmd/kubectl-directpv/utils.go
```diff
@@ -127,3 +102,29 @@ func getInput(msg string) string {
...
+	}
+	printer = printYAML
+	if jsonOutput {
+		printer = printJSON
+	}
+	return nil
+}
+
+func dryRun() bool {
+	return dryRunFlag || yamlOutput || jsonOutput
```
balamurugana, line 129: Why does output mode control dry-run? Many commands would perform actual change and print results as per output mode.

pkg/installer/args.go
Praveenrajmani: installer will just print yaml

cmd/kubectl-directpv/utils.go
Praveenrajmani: `kubectl directpv install -o yaml` is a dry-run mode

Praveenrajmani: we have removed `--dry-run` from install command @balamurugana

cmd/kubectl-directpv/init_config_v1.go
Praveenrajmani: It is used in two places

balamurugana (approved):

PR #710, opened 2023-02-04
Add kubernetes v1.26 support

balamurugana:


balamurugana:
It seems `manusa/actions-setup-minikube` is not yet ready.

PR #715, opened 2023-02-17
Do not rely on the cached objects received in the event handlers

Praveenrajmani:
There is a race identified while updating the CRD objects vs the corresponding cache updated. This makes the values in cache remain outdated / not in sync with the corresponding ETCD values for the object.

As a side-effect and as it cannot verify the outdated values present in the cached objects received in the argument, the handlers could not process the requests like

- mounting the drive on node reboots
- formatting
- releasing
- etc..

as a hack, we ignore the cached objects received in the event handler args, and always the pick the latest object from ETCD.


Praveenrajmani:
Have disabled the `VulnCheck` workflow temporarily as some of the latest GO vulns are not yet fixed in our k8s upstream libraries.

(Note: This PR also updated the deps to their latest released versions)

Praveenrajmani:
Closing this PR as the correct fix is sent in https://github.com/minio/directpv/pull/718

PR #718, opened 2023-02-22
[master] Move to informer based controller for Drive and Volume event handling

Praveenrajmani:
There was a problem in the existing listener implementation, the k8s cache with the indexer isn't updating the changes properly. ie, the cache.object received in the listener side was outdated.. so, the event handlers processing the objects were failing as the data present in these drive and volume objects are wrong/stale..

This issue is seen in reboots because on reboots the drive name could possibly change and such changes were not reflecting on the cache and thus the handlers did not catch up with the latest state of the directcsidrives.. and so, handler could not mount them as the drive values isn't matching the corresponding states probed from the host.

This PR fixes this by completely moving to informer based controller for the object events. This informer is reliable and widely used for controllers in k8s ecosystem.


.golangci.yml
```diff
@@ -27,13 +27,12 @@ linters:
     - misspell
     - govet
     - ineffassign
-    - deadcode
```
Praveenrajmani, line 30: this change is because of

```
WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused.
```

Praveenrajmani:
PTAL @balamurugana @harshavardhana

Praveenrajmani:
this issue is fixed in https://github.com/minio/directpv/releases/tag/v4.0.1

PR #724, opened 2023-03-08
Update the h/w properties on mismatch

Praveenrajmani:
Few OS patches on the nodes can normalize/change the property values. Like,

```
wwid from 'naa.XXX' -> 'XXX'
serial from 'SERIAL NO' -> 'SERIAL_NO'
```

Currently, DirectPV will not update the h/w properties as it is not expected to change. This is not true and these property values can change.

This PR will accept such property changes for "InUse" and "Ready" drives also.

**Symptoms :-**

The symptoms of this issue is that you can see "drive not mounted" errors in workload pods and logs in directpv deamonset pods will report mismatches in h/w properties like wwid, serial no etc.

**Workaround (only applicable for DirectPV v3.2.2) :** To remove the mismatching fields from the corresponding directpvdrive. `kubectl edit directcsidrive <drive-name>`. The drive-name can be taken from `kubectl get directcsivolume <pv-name> -o jsonpath='{.status.drive}'`


Praveenrajmani:
This issue is fixed in https://github.com/minio/directpv/releases/tag/v4.0.1

PR #730, opened 2023-03-23
Add `kubectl directpv drain` command to drain the resources from the node

Praveenrajmani:
This command will be useful to cleanup/drain the resources from a node which was detached or isn't a DirectPV storage node anymore.

This will forcefully remove the following DirectPV resources from the requested node(s)

```
- directpvdrives
- directpvvolumes
- directpvinitrequests
- directpvnode
```

Steps to test 

- Install DirectPV
- Deploy workloads using DirectPV storage class
- Taint a node which isn't needed anymore (simulating a detached node)
- Delete the directpv pod running in that node
- Call `kubectl directpv drain <node>`
- No resources should be listed under that node. ie, check `kubectl directpv info` and `kubectl directpv list` commands to verify that no resources from the detached node are present

### DIFF
--- a/cmd/kubectl-directpv/drain.go
+++ b/cmd/kubectl-directpv/drain.go
@@ -1,0 +1,251 @@
+// This file is part of MinIO DirectPV
+// Copyright (c) 2021, 2022, 2023 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


cmd/kubectl-directpv/drain.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/minio/directpv/pkg/volume"
+	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type drainFunc func(ctx context.Context, nodeName string) error
+
+var drainCmd = &cobra.Command{
+	Use:           "drain [NODE ...]",
```
balamurugana, line 45: 1. Make argument mandatory. `drain <NODE> ...`
2. Add functional tests so that CI will check or make this command hidden

Praveenrajmani:
After some discussions with @balamurugana , we decided to provide a bash script as this command is a cleanup operation and doesn't do much.

PR #739, opened 2023-03-27
Add volume scheduling documentation

### LINKED ISSUE #318: Document volume scheduling (ref: https://github.com/minio/direct-csi/issues/317)
Specifically explain how kubernetes and direct-csi work together to schedule volumes on nodes. Implications of `WaitForFirstConsumer` and topology constraints. 

Also explain how to restrict scheduling to particular nodes

balamurugana:
Fixes #318

### DIFF
--- a/docs/volume-scheduling.md
+++ b/docs/volume-scheduling.md
@@ -1,0 +1,62 @@
+# Volume Scheduling
+
+`DirectPV` comes with storage class named `directpv-min-io` with volume binding mode [WaitForFirstConsumer](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode). This mode delays volume binding and provisioning of a `PersistentVolume` until a `Pod` using the `PersistentVolumeClaim` is created. PersistentVolumes will be selected or provisioned conforming to the topology that is specified by the Pod's scheduling constraints. These include, but are not limited to, resource requirements, node selectors, pod affinity and anti-affinity, and taints and tolerations.
+
+## Drive selection
+
+DirectPV CSI controller selects suitable drive for `CreateVolume` request like below
+1. Filesystem type and/or access-tier in the request is validated. DirectPV supports `xfs` filesystem only.
+2. Each `DirectPVDrive` CRD object is checked whether the requested volume is already present or not. If present, the first drive containing the volume is selected.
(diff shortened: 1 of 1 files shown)


Praveenrajmani:
Do we need a "trouble shooting" section for volume scheduling issues? Like, for eg, if they see issues like "no drive found for requested topology" in the respective pods, we can convey that the the node isn't added (or) the requested size is more etc

ravindk89:
> Do we need a "trouble shooting" section for volume scheduling issues? Like, for eg, if they see issues like "no drive found for requested topology" in the respective pods, we can convey that the the node isn't added (or) the requested size is more etc

It's not a bad idea, but that also largely depends on how common and predictable certain errors are - and how much work it is to maintain that.

But maybe a handful of the most likely suspects (No drives on the node where the pod is scheduled is a good one).
If we go that route though I'd argue we need error codes, not just error messages, to make it easier to do lookups. Otherwise you do string-type matches which can change over time.

balamurugana:
I am thinking docs restructure like below

* installation
  * plugin
    * krew users
    * direct binary users
  * DirectPV
    * default installation
    * customized installation
      * KUBELET_DIR_PATH env var
      * node selector
      * tolerations

* upgrade existing installation
  * plugin
  * DirectPV
    * from v4.x.x
    * from v3.x.x
    * from v2.x.x

* usage guide
  * Add/List/Update/Delete drives
  * Add/List/Update/Delete volumes

* troubleshooting
  * installation issues
    * plugin
    * DirectPV
  * upgrade issues
    * plugin
    * DirectPV
  * Add/List/Update/Delete drives issues
  * Add/List/Update/Delete volumes issues

* architecture
  * plugin
  * DirectPV

* developer guide

I think we would need to address many things than patching this PR

djwfyi:
That doc structure seems reasonable to me, @Praveenrajmani . It covers the major types of docs users would need.

Praveenrajmani:
Can we take this PR and work on the other changes on top? @ravindk89 @djwfyi

Praveenrajmani (approved):
IMO, We can take this in and restructure the docs later

ravindk89:
Entirely reasonable approach. Go for it

PR #752, opened 2023-04-11
Make Node{Stage,Publish}Volume RPCs idempotent on kubelet restarts

balamurugana:
When kubelet is restarted NodeStageVolume/NodePublishVolume RPCs are called with existing staging-target-path and target-path. Previously respective RPCs return errors stating staging-target-path/target-path are already mounted.

This PR fixes the issue.


Praveenrajmani (approved):
Tested and LGTM

PR #753, opened 2023-04-12
Configure XFS max-retry=1 max-timeout=5 for EIO/ENOSPC errors

balamurugana:


harshavardhana (approved):

pkg/xfs/mount_linux.go
```diff
@@ -20,17 +20,36 @@ package xfs
...
+		return err
+	}
+
+	name := path.Base(device)
+	if name == "/" || name == "." {
+		return fmt.Errorf("unable to get device name from device %v", device)
+	}
+
+	if err := os.WriteFile("/sys/fs/xfs/"+name+"/error/metadata/EIO/retry_timeout_seconds", []byte("60"), 0o644); err != nil {
+		return fmt.Errorf("unable to set EIO retry_timeout_seconds for device %v; %w", name, err)
```
harshavardhana, line 45: Actually even 30s is more than enough @balamurugana

harshavardhana:
Oh btw @balamurugana rhel7 doesn't have these files.

harshavardhana (changes requested):

balamurugana:
> Oh btw @balamurugana rhel7 doesn't have these files.
> 
> It's only on rhel8 and above..

@harshavardhana rhel 7 has the support
```sh
[root@master ~]# cat /etc/redhat-release 
CentOS Linux release 7.9.2009 (Core)

[root@master ~]# uname -a
Linux master 3.10.0-1160.88.1.el7.x86_64 #1 SMP Tue Mar 7 15:41:52 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

[root@master ~]# ls -1 /sys/fs/xfs/vdb/error/metadata/{EIO,ENOSPC}/retry_timeout_seconds
/sys/fs/xfs/vdb/error/metadata/EIO/retry_timeout_seconds
/sys/fs/xfs/vdb/error/metadata/ENOSPC/retry_timeout_seconds

[root@master ~]# cat /sys/fs/xfs/vdb/error/metadata/EIO/retry_timeout_seconds /sys/fs/xfs/vdb/error/metadata/ENOSPC/retry_timeout_seconds
-1
-1

[root@master ~]# echo 60 > /sys/fs/xfs/vdb/error/metadata/EIO/retry_timeout_seconds; echo 60 > /sys/fs/xfs/vdb/error/metadata/ENOSPC/retry_timeout_seconds 
[root@master ~]# cat /sys/fs/xfs/vdb/error/metadata/EIO/retry_timeout_seconds /sys/fs/xfs/vdb/error/metadata/ENOSPC/retry_timeout_seconds
60
60
[root@master ~]# 
```

pkg/xfs/mount_linux.go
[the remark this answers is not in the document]
harshavardhana: These files are missing because kernel doesn't support until RHEL8 - we can't afford for DirectPV to throw errors on existing deployment.

harshavardhana: Also we need to set this to 1 for max_retries, we really don't want any retry via XFS beyond atleast 1 entry.

And perhaps we should reduce the retry only up-to 5s or worst 10s.

harshavardhana (changes requested):

balamurugana:
Functional tests on rhel9 fails

Praveenrajmani (approved):

Praveenrajmani (approved):

harshavardhana (approved):

PR #754, opened 2023-04-12
Remove /proc/mounts fallback support.

balamurugana:
Previously if `/proc/1/mountinfo` does not exist, `/proc/mounts` is used to probe mounts as a fallback. As /proc/mounts has lack of information about bind mounts, the fallback support is removed.


pkg/sys/mount_linux.go
```diff
@@ -93,91 +92,14 @@ func parseProc1Mountinfo(r io.Reader) (mountPointMap, deviceMap, majorMinorMap,
...
-			majorMinorMap[majorMinor] = make(utils.StringSet)
-		}
-		majorMinorMap[majorMinor].Set(device)
-	}
-
-	return
-}
-
-func getMounts(includeMajorMinorMap bool) (mountPointMap, deviceMap, majorMinorMap, rootMountPointMap map[string]utils.StringSet, err error) {
+func getMounts(_ bool) (mountPointMap, deviceMap, majorMinorMap, rootMountPointMap map[string]utils.StringSet, err error) {
```
Praveenrajmani, line 95: can we remove this arg if it is not used @balamurugana

balamurugana: We will remove in future so that this PR can be reverted easily if needed

Praveenrajmani (approved):

PR #763, opened 2023-05-04
Log and continue on XFS probe errors during startup

Praveenrajmani:
Currently, when any device on the host fails while probing XFS, we error out which may lead to pod crashloops. We can ignore such device probe errors and can log and continue.

### DIFF
--- a/pkg/device/sync.go
+++ b/pkg/device/sync.go
@@ -57,7 +57,7 @@ func probeDeviceMap() (map[string][]device, error) {
 		fsuuid, label, totalCapacity, freeCapacity, err := xfs.Probe(utils.AddDevPrefix(dev.Name))
 		if err != nil {
 			if !errors.Is(err, xfs.ErrFSNotFound) {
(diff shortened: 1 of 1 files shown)


balamurugana (approved):

pkg/device/sync.go
```diff
@@ -57,7 +57,7 @@ func probeDeviceMap() (map[string][]device, error) {
 		fsuuid, label, totalCapacity, freeCapacity, err := xfs.Probe(utils.AddDevPrefix(dev.Name))
 		if err != nil {
 			if !errors.Is(err, xfs.ErrFSNotFound) {
-				return nil, err
+				klog.ErrorS(err, "unable to probe the device", "Device", dev.Name)
```
balamurugana, line 60: You could specifically say "unable to probe XFS filesystem"

balamurugana (changes requested):

balamurugana (approved):

PR #764, opened 2023-05-04
fix: Honour 'ctrl+c' in the UI progress model

Praveenrajmani:
Also, fixes a minor bug in deployment deletion

### DIFF
--- a/cmd/kubectl-directpv/progress_model.go
+++ b/cmd/kubectl-directpv/progress_model.go
@@ -82,6 +82,11 @@ func (m progressModel) Init() tea.Cmd {
 
 func (m progressModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
+	case tea.KeyMsg:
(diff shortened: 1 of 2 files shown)


cmd/kubectl-directpv/progress_model.go
```diff
@@ -82,6 +82,14 @@ func (m progressModel) Init() tea.Cmd {
 
 func (m progressModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		switch msg.String() {
+		case "ctrl+c":
+			var cmds []tea.Cmd
+			cmds = append(cmds, tea.Sequence(finalPause(), tea.Quit))
+			return m, tea.Batch(cmds...)
+		}
```
balamurugana, line 90: You could do just
```go
if msg.String() == "ctrl+c" {
    return m, tea.Batch(tea.Sequence(finalPause(), tea.Quit))
}
return m, nil
```

balamurugana (approved):

PR #768, opened 2023-05-17
Add FAQ section in docs

Praveenrajmani:


balamurugana (approved):

docs/faq.md
```diff
@@ -0,0 +1,54 @@
+FAQ
+----
+
+#### What type of disks are recommended for DirectPV?
+
+DirectPV is specifically meant for [Direct Attached Storage](https://en.wikipedia.org/wiki/Direct-attached_storage) such as hard drives, solid-state drives and JBODs etc. Whereas, SAN and NAS based storage options are not recommended as it involves extra network hops in the data path. This may lead to poor perfomance and increased complexity.
```
ravindk89, line 6: Shall we be more forceful here?

```suggestion
DirectPV is specifically meant for [Direct Attached Storage](https://en.wikipedia.org/wiki/Direct-attached_storage) such as hard drives, solid-state drives and JBODs etc. 

Avoid using DirectPV with SAN and NAS based storage options, as they inherently involve extra network hops in the data path. This leads to poor performance and increased complexity.
```

docs/faq.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### How DirectPV is different from LocalPV and HostPath?
+
+Hostpath volumes are ephemeral and are tied to the lifecyle of pods. ie, the volumes are lost when a pod is restarted or deleted. Whereas, DirectPV volumes are persistent and can sustain node and pod reboots. And the lifecycle of a DirectPV volume is managed by [Persistent Volume Claims](https://kubernetes.io/docs/concepts/storage/persistent-volumes/).
```
ravindk89, line 10:
```suggestion
Hostpath volumes are ephemeral and are tied to the lifecyle of pods. Hostpath volumes are lost when a pod is restarted or deleted, resulting in the loss of any stored data. 

DirectPV volumes are persistent through node and pod reboots. The lifecycle of a DirectPV volume is managed by the associated [Persistent Volume Claims](https://kubernetes.io/docs/concepts/storage/persistent-volumes/).
```

docs/faq.md
```diff
@@ -0,0 +1,54 @@
...
+
+LocalPV is statically provisioned where the volumes need to be pre-provisioned before the workloads using them are created. DirectPV is similar to localPV but it supports dynamic provisioning.
```
ravindk89, line 12:
```suggestion
LocalPVs are statically provisioned from a local storage resource on the nodes where persistent volumes are required, and must be created prior to the workloads using them. User management is required down to creating the actual object resource for the local PV.

DirectPV also creates statically provisioned volume resources, but does not require user intervention for the creation of the object resource. Instead, DirectPV dynamically provisions the persistent volume in response to a PVC requesting a DirectPV storage volume. This significantly reduces complexity of management.

```

docs/faq.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### How are the disks selected for a Pod?
+
+DirectPV selects the disks local to the node where the pod is scheduled. This provides direct access for the pods to the disks. DirectPV runs a selection algorithm to choose a disk for a volume creation request and then creates a sub-directory for the volume with quota set on the sub-directory for the requested volume size and publishes the volume to the pod. To know more on the selection algorithm, please refer [here](./volume-scheduling.md).
```
ravindk89, line 16:
```suggestion
DirectPV operates only on those disks which it [explicitly manages](./cli.md#initialize-the-available-drives-present-in-the-cluster)

1. DirectPV selects managed disks local to the node where the pod is scheduled. This provides direct access for the pods to the disks. 

2. DirectPV runs a selection algorithm to choose a disk for a volume creation request

 3. DirectPV then creates a sub-directory for the volume with quota set on the sub-directory for the requested volume size
 
 4. DirectPV publishes the volume to the pod. 
 
 To know more on the selection algorithm, please refer [here](./volume-scheduling.md).
```

docs/faq.md
```diff
@@ -0,0 +1,54 @@
...
+A drive must meet the following requirements for it to be used for initialization
+
+- The size of the drive should not be less then 512MiB.
+- The drive should not be hidden. (Check /sys/class/block/<device>/hidden, It should not be "1")
+- The drive should not be a read-only.
+- The drive should not be partitioned.
+- The drive should not be held by other devices. (Check /sys/class/block/<device>/holders, It should not have any entries)
+- The drive should not be mounted.
+- The drive should not have swap-on enabled.
+- The drive should not be a CDROM.
```
ravindk89, line 33: Replace "should not" with "must not" if htese are requirements.

ravindk89: Also wondering if we should state the drive must not be an LVM, LUN, or similar virtualized drive.

ravindk89 (changes requested):
A few suggestions for slightly tightening a few things.

docs/faq.md
Praveenrajmani: we do support LVMs,LUKS and LUNs @ravindk89 It still works with DirectPV

Just that we don't recommend them for production setups.

balamurugana: Technically we support any block device, but we do not recommend network drives, LVMs, etc. Hence LUKS on top of DAS is right way to use, we support it.

ravindk89: In that case maybe below this block state

"DirectPV can support any storage resource presenting itself as a block device. While this includes LVMs, LUKs, LUNs, and similar virtualized drives, we do not recommend using DirectPV with those in production environments without SUBNET support."

Or something to that effect.

ravindk89 (approved):
LGTM regardless of how last comment is resolved.

docs/faq.md
Praveenrajmani: Added this @ravindk89 

"DirectPV can support any storage resource presenting itself as a block device. While this includes LVMs, LUKs, LUNs, and similar virtualized drives, we do not recommend using DirectPV with those in production environments."

balamurugana (approved):

PR #772, opened 2023-05-18
Add supported OC version for Red Hat CSI Certification

cniackz:
### Objective:

To set the supported versions of OpenShift for DirectPV.

### Why this is needed?

* Since some old versions of OpenShift has been deprecated, we should specify the range that is supported for OLM tests to pass under these recent versions, from 4.8 onwards. 

### Documentation:

* https://docs.openshift.com/container-platform/4.8/operators/operator_sdk/osdk-working-bundle-images.html


balamurugana (approved):

Praveenrajmani:
can you fix the conflict here @cniackz

Praveenrajmani (approved):

PR #780, opened 2023-05-26
Add declarative installation

balamurugana:


docs/tools/install.sh
```diff
@@ -0,0 +1,228 @@
...
+    fi
+
+    echo "* Generate DirectPV manifests to be saved at directpv.yaml"
+    if [[ "${legacy}" -eq 0 ]]; then
+        run_plugin_cmd install --declarative "${directpv_args[@]}" > directpv.yaml
+    else
+        run_plugin_cmd install --legacy --declarative "${directpv_args[@]}" > directpv.yaml
+    fi
+
+    if [ "${apply_flag}" -eq 1 ]; then
```
Praveenrajmani, line 201:
```suggestion
    if [ "${apply_flag}" -eq 0 ]; then
```

Currently, it applies even if `apply` flag is not provided

docs/tools/install.sh
```diff
@@ -0,0 +1,228 @@
...
+
+EXAMPLES:
+  # Generate DirectPV manifests.
+  $ install.sh
+
+  # Generate DirectPV manifests with private registry.
+  $ install.sh --registry private-registry.io --org org-name
+
+  # Generate DirectPV manifests with node-selector.
+  $ install.sh --node-selector node-label-key=node-label-value
```
Praveenrajmani, line 145: Is there a way to test local dev images? can we also add `--image`?

pkg/installer/args.go
```diff
@@ -34,40 +37,37 @@ const (
...
+	Tolerations      []corev1.Toleration
+	SeccompProfile   string
+	AppArmorProfile  string
+	Quiet            bool
+	KubeVersion      *version.Version
+	Legacy           bool
+	ObjectWriter     io.Writer
+	DryRun           bool
+	Declarative      bool
+	ObjectConverter  func(runtime.Object) ([]byte, error)
```
Praveenrajmani, line 56: can we call it as Marshaller/ObjectMarshaller?

docs/tools/install.sh
balamurugana: You can pass whatever `kubectl-directpv install` command has. These examples are to indicate how to use `install` command's arguments.

docs/tools/install.sh
balamurugana: done

pkg/installer/args.go
balamurugana: done

docs/tools/install.sh
Praveenrajmani: this is not an issue also, we can edit the image manually in the manifest. just that we can't use the script

Praveenrajmani (approved):

docs/tools/install.sh
balamurugana: Your expectation is wrong. The image you are expecting has to be in the registry or in the local registry. It is a bad idea to modify the manifests here.

Praveenrajmani: I meant we can do `kubectl directpv install --declarative > manifest.yaml` and then editing the `image` field to our dev image.

I see a point that `dev` images will not be of real use here. So, there is not need to use this script for development.

PR #782, opened 2023-05-29
fix: `kubectl directpv move` to handle empty volumes in source drive

Praveenrajmani:
This patch includes the following fixes

- Checks the finalizers and exits if empty list is found in 'kubectl directpv move'
- Use drive ids instead of drive names in 'docs/tools/replace.sh'. As the drive names are not reliable and guaranteed.

### DIFF
--- a/cmd/kubectl-directpv/move.go
+++ b/cmd/kubectl-directpv/move.go
@@ -83,9 +83,15 @@ func moveMain(ctx context.Context, src, dest string) {
 		os.Exit(1)
 	}
 
+	sourceVolumeNames := srcDrive.GetVolumes()
+	if len(sourceVolumeNames) == 0 {
+		utils.Eprintf(quietFlag, false, "No volumes found in source drive %v\n", src)
+		return
+	}
+
 	var requiredCapacity int64
 	var volumes []types.Volume
-	for result := range volume.NewLister().VolumeNameSelector(srcDrive.GetVolumes()).List(ctx) {
+	for result := range volume.NewLister().VolumeNameSelector(sourceVolumeNames).List(ctx) {
 		if result.Err != nil {
 			utils.Eprintf(quietFlag, true, "%v\n", result.Err)
 			os.Exit(1)
--- a/docs/tools/replace.sh
+++ b/docs/tools/replace.sh
@@ -22,11 +22,15 @@
 
 set -e
 
-# usage: get_drive_id <node> <drive-name>
-function get_drive_id() {
+ME=$(basename "$0"); export ME
+
+export drive_id=""
+
+# usage: get_drive_ids <node> <drive-name>
+function get_drive_ids() {
     kubectl get directpvdrives \
             --selector="directpv.min.io/node==${1},directpv.min.io/drive-name==${2}" \
-            -o go-template='{{range .items}}{{.metadata.name}}{{end}}'
+            -o go-template='{{range .items}}{{.metadata.name}} {{end}}'
 }
 
 # usage: get_volumes <drive-id>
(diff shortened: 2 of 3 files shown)


cmd/kubectl-directpv/move.go
```diff
@@ -101,7 +107,7 @@ func moveMain(ctx context.Context, src, dest string) {
 	}
 
 	if len(volumes) == 0 {
-		utils.Eprintf(quietFlag, false, "No volumes found in source drive %v\n", src)
+		utils.Eprintf(quietFlag, false, "No %s volumes found in source drive %v\n", consts.AppPrettyName, src)
```
balamurugana, line 110: `DirectPV` is obvious name.

docs/tools/replace.sh
```diff
@@ -36,23 +36,39 @@ function get_volumes() {
...
+}
+
+# usage: is_uuid input
+function is_uuid() {
+    if [[ "${1}" =~ ^\{?[A-F0-9a-f]{8}-[A-F0-9a-f]{4}-[A-F0-9a-f]{4}-[A-F0-9a-f]{4}-[A-F0-9a-f]{12}\}?$ ]]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+}
```
balamurugana, line 67: Just do below. IMO we don't need to allow `{A-F}` as DirectPV programmatically doesn't generate in that pattern
```bash
function is_uuid() {
    [[ "$1" =~ ^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$ ]]
}
```

docs/tools/replace.sh
```diff
@@ -36,23 +36,39 @@ function get_volumes() {
...
+
 function init() {
-    if [[ $# -eq 4 ]]; then
-        echo "usage: replace.sh <NODE> <SRC-DRIVE> <DEST-DRIVE>"
+    if [[ $# -ne 3 ]]; then
```
balamurugana, line 70: the check should be like `if [[ $# -lt 3 || $# -gt 4 ]]; then`

docs/tools/replace.sh
```diff
@@ -70,21 +86,73 @@ function init() {
...
-
-    # Get source drive ID
-    src_drive_id=$(get_drive_id "${node}" "${src_drive}")
-    if [ -z "${src_drive_id}" ]; then
-        echo "source drive ${src_drive} on node ${node} not found"
+    src_drive="${1#/dev/}"
+    dest_drive="${2#/dev/}"
+    node="${3}"
+
+    if [ "$(is_uuid "${src_drive}")" == "false" ]
```
balamurugana, line 93: just do `if ! is_uuid "${src_drive}"; then`

docs/tools/replace.sh
```diff
@@ -70,21 +82,69 @@ function init() {
...
+    dest_drive="${2#/dev/}"
+    node="${3}"
+
+    if ! is_uuid "${src_drive}"; then
+        if [ -z "${node}" ]; then
+            echo "node argument missing; please check the usage"
+            exit 1
+        fi
+        # Get source drive ID
+        src_drive_ids=$(get_drive_ids "${node}" "${src_drive}")
```
balamurugana, line 95: Arrays should be initiated like `src_drive_ids=( $(get_drive_ids ...) )`

docs/tools/replace.sh
```diff
@@ -70,21 +82,69 @@ function init() {
...
+        if [ -z "${src_drive_ids}" ]; then
```
balamurugana, line 96: The condition is like below
```sh
if [ "${#src_drive_ids[@]}" -eq 0 ]; then
    # source drive not found
    exit 255
fi
if [ "${#src_drive_ids[@]}" -gt 1 ]; then
    # too many source drives by name found
    exit 255
fi
src_drive_id="${src_drive_ids[0]}"
```

docs/tools/replace.sh
```diff
@@ -70,21 +82,69 @@ function init() {
...
+        if [ "${#src_drive_ids[@]}" -gt 1 ]; then
+            echo "duplicate drive ids found for '${src_drive}'"
+            exit 1
+        fi
+        src_drive_id="$(echo "${src_drive_ids[0]}" | xargs)"
+    else
+        src_drive_id="${src_drive}"
+    fi
+
+    if ! is_uuid "${dest_drive}"; then
```
balamurugana, line 109: same as above

docs/tools/replace.sh
```diff
@@ -70,21 +82,69 @@ function init() {
...
-    dest_drive="${3#/dev/}"
-
-    # Get source drive ID
-    src_drive_id=$(get_drive_id "${node}" "${src_drive}")
-    if [ -z "${src_drive_id}" ]; then
-        echo "source drive ${src_drive} on node ${node} not found"
+    src_drive="${1#/dev/}"
+    dest_drive="${2#/dev/}"
+    node="${3}"
+
```
balamurugana, line 88: You could error out if `src_drive == dest_drive` here. Also error out if `node` argument is missing if `src_drive`/`dest_drive` arguments are drive names.

docs/tools/replace.sh
```diff
@@ -70,21 +101,74 @@ function init() {
...
+        exit 255
+    fi
+
+    if ! is_uuid "${src_drive}"; then
+        if [ -z "${node}" ]; then
+            echo "node argument must be provided for drive name"
+            exit 255
+        fi
+        # Get source drive ID
+        mapfile -t src_drive_ids < <(get_drive_ids "${node}" "${src_drive}")
```
balamurugana, line 119: 1. use `src_drive_ids=( $(get_drive_ids "${node}" "${src_drive}") )` and ignore ` # shellcheck disable=SC2207`. This way you could use `dest_drive_id="${dest_drive_ids[0]}"`
2. This code block repeats for destination drive as well. You could have a common function.

docs/tools/replace.sh
```diff
@@ -36,25 +41,73 @@ function get_volumes() {
...
+        return
+    fi
+    # shellcheck disable=SC2207
+    drive_ids=( $(get_drive_ids "${1}" "${2}") )
+    if [ "${#drive_ids[@]}" -eq 0 ]; then
+        id_err="drive '${2}' on node '${1}' not found"
+        return
+    fi
+    if [ "${#drive_ids[@]}" -gt 1 ]; then
+        id_err="duplicate drive ids found for '${2}'"
```
balamurugana, line 83: You exit right here after printing error.  There is no need to store. If you feel function name would confuse, use `must_get_drive_id`

docs/tools/replace.sh
```diff
@@ -70,21 +123,56 @@ function init() {
...
+        echo "the source and destination drives are same"
+        exit 255
+    fi
+
+    if ! is_uuid "${src_drive}"; then
+        get_drive_id "${node}" "${src_drive}"
+        if [ "${id_err}" != "" ]; then
+            echo "$id_err"
+            exit 255
+        fi
```
balamurugana, line 140: You can remove this.

docs/tools/replace.sh
```diff
@@ -107,8 +195,7 @@ function main() {
         pod_namespace=$(get_pod_namespace "${volume}")
 
         if ! kubectl delete pod "${pod_name}" --namespace "${pod_namespace}"; then
-            echo "unable to delete pod ${pod_name} using volume ${volume}"
-            exit 1
+            echo "unable to delete pod '${pod_name}' using volume '${volume}'"
```
balamurugana, line 198: You could change to `echo "unable to delete pod '${pod_name}' using volume '${volume}'; please restart the pod manually`

balamurugana (approved):

PR #799, opened 2023-06-10
Ignore sysfs file read errors

harshavardhana:


Praveenrajmani:
we can set `false` to the readFirstLine function here - https://github.com/minio/directpv/blob/c0ae169c3b169219a4672bb753872c35381410b6/pkg/device/sysfs_linux.go#L75

balamurugana:
> we can set `false` to the readFirstLine function here -
> 
> https://github.com/minio/directpv/blob/c0ae169c3b169219a4672bb753872c35381410b6/pkg/device/sysfs_linux.go#L75

@harshavardhana @Praveenrajmani Do you guys see `sysfs` devices with no `size` entry?

Praveenrajmani:
yes @balamurugana 

portworx volumes are missing this

harshavardhana:
> > we can set `false` to the readFirstLine function here -
> > https://github.com/minio/directpv/blob/c0ae169c3b169219a4672bb753872c35381410b6/pkg/device/sysfs_linux.go#L75
> 
> @harshavardhana @Praveenrajmani Do you guys see `sysfs` devices with no `size` entry?

Yes portworx devices seems to have it missing, an another CSI product.

harshavardhana:
> we can set `false` to the readFirstLine function here -
> 
> https://github.com/minio/directpv/blob/c0ae169c3b169219a4672bb753872c35381410b6/pkg/device/sysfs_linux.go#L75

We can simply ignore all devices that do not comply with what we want. I am not big fan of failing pods with no easy to understand errors.

Unless we want to log why we are ignoring the drives? That is okay too.

balamurugana:
> 

@harshavardhana You could pass `false` at https://github.com/minio/directpv/blob/master/pkg/device/sysfs_linux.go#L75 which fixes the issue.

harshavardhana:
> @harshavardhana You could pass `false` at https://github.com/minio/directpv/blob/master/pkg/device/sysfs_linux.go#L75 which fixes the issue.

No @balamurugana that is not a complete fix. We should ignore all devices that do not comply with Disk metadata that we want.

It is not just about `/size` there are other files that may be missing. There is no document that indicates that these files "must" be present or they are mandatory.

The current PR simply addresses this more comprehensively. For example, portworx has a couple of other files missing as well. If we only fixed `/size` alone is not sufficient.

IMO we shouldn't crash our pods unnecessarily. We should work when we have other good drives we can manage.

This is a support nightmare and also a bad POC experience. We should be able to deploy DirectPV without @balamurugana or @Praveenrajmani's involvement.

pkg/device/probe_linux.go
```diff
@@ -106,6 +106,10 @@ func probe() (devices []Device, err error) {
 	for name, udevData := range udevDataMap {
 		device, err := newDevice(deviceMountMap, majorMinorMap, cdroms, swaps, name, deviceMap[name], udevData)
 		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
```
balamurugana, line 109: This check is not needed anymore.

pkg/device/probe_linux.go
```diff
@@ -142,11 +146,19 @@ func probeDevices(majorMinor ...string) (devices []Device, err error) {
 
 		name, err := getDeviceName(majorMinor[i])
 		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
```
balamurugana, line 149: This check is not needed anymore.

pkg/device/probe_linux.go
```diff
@@ -142,11 +146,19 @@ func probeDevices(majorMinor ...string) (devices []Device, err error) {
...
+				continue
+			}
+
 			return nil, err
 		}
 
 		device, err := newDevice(deviceMountMap, majorMinorMap, cdroms, swaps, name, majorMinor[i], udevData)
 		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
```
balamurugana, line 158: This check is not needed anymore.

balamurugana (approved):

Praveenrajmani (approved):

PR #810, opened 2023-07-07
Add macOS and OS architecture support in olm.sh

cniackz:
### Objective:

To allow different OSes and Architectures in olm.sh

### Reasoning:

* I am using a MacBook Pro, meaning:

```sh
$ uname
Darwin
```

and

```sh
$ uname -m
arm64
```

### Related:

* This will help me create new bundle for certification.

### DIFF
--- a/olm.sh
+++ b/olm.sh
@@ -33,8 +33,8 @@ function get_yq() {
 }
 
 function get_kubectl_directpv() {
-    [ -f "kubectl-directpv_${BUILD_VERSION}_linux_amd64" ] && return 0
-    curl --silent --location --insecure --fail --output "${KUBECTL_DIRECTPV}" "https://github.com/minio/directpv/releases/download/v${BUILD_VERSION}/kubectl-directpv_${BUILD_VERSION}_linux_amd64"
+    [ -f "${KUBECTL_DIRECTPV}" ] && return 0
+    curl --silent --location --insecure --fail --output "${KUBECTL_DIRECTPV}" "https://github.com/minio/directpv/releases/download/v${BUILD_VERSION}/${KUBECTL_DIRECTPV:2}"
     chmod a+x "${KUBECTL_DIRECTPV}"
 }
 
@@ -53,7 +53,7 @@ EOF
     # assign after trimming 'v'
     BUILD_VERSION="${1/v/}"
     IMAGE_HASH="$2"
-    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_linux_amd64"
+    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)"
 
     echo "Downloading required tools"
 


olm.sh
```diff
@@ -53,7 +53,9 @@ EOF
     # assign after trimming 'v'
     BUILD_VERSION="${1/v/}"
     IMAGE_HASH="$2"
-    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_linux_amd64"
+    OS=$(uname | tr '[:upper:]' '[:lower:]')
+    ARCH=$(uname -m)
```
balamurugana, line 57: This script is meant for us i.e. MinIO maintainers. I don't think this level of support is needed. Supporting `linux_amd64` is good enough.

If you think we still need to support `kubectl-directpv` release on `linux`, `darwin` and `windows` on `amd64`, `arm64` and `ppc64le`, we would need to fix it properly.

Instead of adding such complicated support, I would rather error out if this script is run on non-linux on non-amd64.

balamurugana (changes requested):

olm.sh
cniackz: @balamurugana I totally understand your point but we are not using the same architecture. I am using a MacBook Pro and this change will be very useful for me personally. Other guys like @dvaldivia, @pjuarezd  and @reivaj05 are also using Darwin systems; I believe this will benefit us since we are running and maintaining code under different architectures and Operating Sytems.

balamurugana: Got it. The fix would be like below
```diff
diff --git a/olm.sh b/olm.sh
index 67a13d93..c44114c0 100755
--- a/olm.sh
+++ b/olm.sh
@@ -33,8 +33,8 @@ function get_yq() {
 }
 
 function get_kubectl_directpv() {
-    [ -f "kubectl-directpv_${BUILD_VERSION}_linux_amd64" ] && return 0
-    curl --silent --location --insecure --fail --output "${KUBECTL_DIRECTPV}" "https://github.com/minio/directpv/releases/download/v${BUILD_VERSION}/kubectl-directpv_${BUILD_VERSION}_linux_amd64"
+    [ -f "${KUBECTL_DIRECTPV}" ] && return 0
+    curl --silent --location --insecure --fail --output "${KUBECTL_DIRECTPV}" "https://github.com/minio/directpv/releases/download/v${BUILD_VERSION}/${KUBECTL_DIRECTPV}"
     chmod a+x "${KUBECTL_DIRECTPV}"
 }
 
@@ -53,7 +53,7 @@ EOF
     # assign after trimming 'v'
     BUILD_VERSION="${1/v/}"
     IMAGE_HASH="$2"
-    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_linux_amd64"
+    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)"
 
     echo "Downloading required tools"
 
```

`uname` outputs different values with respect to release binary names.

balamurugana (approved):

pjuarezd (approved):

PR #819, opened 2023-07-24
Implement volume health monitoring feature to detect abnormal volumes

Praveenrajmani:
This is a [CSI feature](https://kubernetes.io/docs/concepts/storage/volume-health-monitoring/) which monitors the volume health, identifies and reports the abnormal volumes as events to the respective pods and PVCs.

Steps to test :


Praveenrajmani:
For this PR, we need  `registry.k8s.io/sig-storage/csi-external-health-monitor-controller:v0.9.0` to be pushed to `quay.io/minio/`

pkg/consts/consts.go
```diff
@@ -94,4 +94,7 @@ const (
 
 	LegacyNodeServerName       = "legacy-node-server"
 	LegacyControllerServerName = "legacy-controller"
+
+	// Volume Health Monitor
+	VolumeHealthMonitorIntervalInDuration = "1m"
```
Praveenrajmani, line 99: will set it to `15m` by default and make it configurable

pkg/volume/volume-health.go
```diff
@@ -0,0 +1,122 @@
...
+func checkVolumeHealth(ctx context.Context, volumeName string, getVolumeMounts func(string) utils.StringSet) error {
+	volume, err := client.VolumeClient().Get(
+		ctx, volumeName, metav1.GetOptions{TypeMeta: types.NewVolumeTypeMeta()},
+	)
+	if err != nil {
+		return err
+	}
+	var hasError bool
+	hasError, err = checkVolumeMounts(ctx, volume, getVolumeMounts)
+	if err != nil || hasError {
```
Praveenrajmani, line 75: we can use typed errors here. ie. `errMountDoNotExist`.

pkg/k8s/k8s.go
```diff
@@ -162,6 +196,22 @@ func BoolToConditionStatus(val bool) metav1.ConditionStatus {
 	return metav1.ConditionFalse
 }
 
+// ConditionStatusToBool converts condition status to bool.
+func ConditionStatusToBool(status metav1.ConditionStatus) bool {
+	return status == metav1.ConditionTrue
+}
+
+// ResetConditionIfMatches resets the condition values to default if the type, reason and message matches.
+func ResetConditionIfMatches(conditions []metav1.Condition, ctype string, reason, message, newReason string) {
```
balamurugana, line 205: This function is not needed. Use `UpdateCondition()`

pkg/csi/utils/utils.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	directpvtypes "github.com/minio/directpv/pkg/apis/directpv.min.io/types"
+	"github.com/minio/directpv/pkg/k8s"
+	"github.com/minio/directpv/pkg/types"
+)
+
+// GetCSIVolumeCondition returns the CSI volume condition
+func GetCSIVolumeCondition(volume *types.Volume) *csi.VolumeCondition {
```
balamurugana, line 27: Looks like this function is not used anywhere. If you use this in this PR, make this function as method to `types.Volume` and remove this file

docs/monitoring.md
```diff
@@ -79,3 +79,21 @@ scrape_configs:
...
+To enable volume health monitoring, Install directpv with `--enable-volume-health-monitoring` flag.
+
+```sh
+kubectl directpv install --enable-volume-health-monitoring
+```
+
+For private registries, please note that the following image is required for enabling volume health monitoring
+
+```
+quay.io/minio/csi-external-health-monitor-controller:v0.10.0
```
balamurugana, line 98: 1. You would need to update https://github.com/minio/directpv/blob/master/docs/tools/push-images.sh instead of documenting here
2. You would need to update https://github.com/minio/directpv/blob/master/docs/command-reference.md and https://github.com/minio/directpv/blob/master/docs/installation.md#directpv-csi-driver-installation

docs/monitoring.md
```diff
@@ -79,3 +79,21 @@ scrape_configs:
     action: replace
     target_label: kubernetes_name
 ```
+
+# Volume health monitoring
```
balamurugana, line 83: You could have the doc like
```
# Monitoring volume health
DirectPV monitors volume health and reports abnormalities as event to PVCs and Pods. Currently volume mounts are monitored and more checks will be added later. This feature uses [Volume health monitoring CSI feature](https://kubernetes.io/docs/concepts/storage/volume-health-monitoring/).

To enable volume health monitoring, pass `--enable-volume-health-monitoring` flag to installation and `CSIVolumeHealth` feature gate must be enabled.
```

pkg/installer/args.go
```diff
@@ -47,30 +47,34 @@ const (
 	// csiResizerImage = csi-resizer:v1.8.0
 	csiResizerImage          = "csi-resizer@sha256:819f68a4daf75acec336302843f303cf360d4941249f9f5019ffbb690c8ac7c0"
 	openshiftCSIResizerImage = "registry.redhat.io/openshift4/ose-csi-external-resizer-rhel8@sha256:837b32a0c432123e2605ad6d029e7f3b4489d9c52a9d272c7a133d41ad10db87"
+
+	// csiHealthMonitorImage = csi-external-health-monitor-controller:v0.10.0
+	csiHealthMonitorImage = "registry.k8s.io/sig-storage/csi-external-health-monitor-controller:v0.10.0"
```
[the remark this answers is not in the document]
Praveenrajmani: will use the `quay.io/minio/` image after pushing to the registry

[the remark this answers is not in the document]
Praveenrajmani: we don't have any equivalent image for this available in redhat openshift

pkg/apis/directpv.min.io/types/types.go
```diff
@@ -120,23 +120,28 @@ type VolumeConditionType string
...
 
 // VolumeConditionReason denotes volume reason. Allows maximum upto 1024 chars.
 type VolumeConditionReason string
 
 // Enum values of VolumeConditionReason type.
 const (
-	VolumeConditionReasonDriveLost VolumeConditionReason = "DriveLost"
+	VolumeConditionReasonDriveLost  VolumeConditionReason = "DriveLost"
+	VolumeConditionReasonNotMounted VolumeConditionReason = "NotMounted"
+	VolumeConditionReasonNoError    VolumeConditionReason = "NoError"
```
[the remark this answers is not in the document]
Praveenrajmani: The reason in a condition cannot be empty. So, "normal" volumes will have the reason "NoError" for abnormal volumes.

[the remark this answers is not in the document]
Praveenrajmani: > The reason in a condition cannot be empty

docs/monitoring.md
Praveenrajmani: Volume health monitoring is not enabled by default, so, for a fresh installation we don't need this. Users need this image only when they enable volume health monitoring

docs/monitoring.md
Praveenrajmani: I have also added the following because this image is required only when volume health monitor is enabled.

```
For private registries, please note that the following image is required for enabling volume health monitoring

quay.io/minio/csi-external-health-monitor-controller:v0.10.0

```

pkg/k8s/k8s.go
Praveenrajmani: This function is required because we need to reset the condition to default only if all the condition value matches. Unlike `UpdateCondition()` where only condition type is checked

pkg/installer/args.go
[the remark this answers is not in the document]
balamurugana: > we don't have any equivalent image for this available in redhat openshift

Then we are stuck to certify DirectPV for Openshift certification. We would need to build `csi-external-health-monitor-controller` our own to certify it for openshift

pkg/apis/directpv.min.io/types/types.go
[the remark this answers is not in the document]
balamurugana: Kubernetes uses positive reasons in conditions, but we use negative reasons. It looks odd for `Error` type having `NoError` reason. Etiher you could deal with `Status == metav1.ConditionTrue` as enabled condition or have positive reasons.

docs/monitoring.md
balamurugana: Users make common mistakes to miss pulling this image when enabling it. There is no problem to have unused image of `csi-external-health-monitor-controller`. When health monitoring is enabled, it works seamlessly than failures.

pkg/installer/args.go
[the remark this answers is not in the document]
Praveenrajmani: Have sent a PR to use a script to download and release the image which can comply with RedHat image certification testcases.

PR #821, opened 2023-08-04
Add Helm chart for DirectPV operator

cniackz:
### Objective:

Add `Helm Chart` for DirectPV

### Reasoning:

This will be needed for our DirectPV Helm Operator that will be used for the certification. When initializing the Operator, we have a flag where this folder goes: `--helm-chart` then Operator will use that helm chart for reconciling.

### Documentation:

* Thank you @balamurugana for providing the notes and for finding this method!

* https://github.com/cniackz/public/wiki/Helm-Operator-for-DirectPV

### DIFF
--- a/helm.sh
+++ b/helm.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+# This file is part of MinIO DirectPV
+# Copyright (c) 2023 MinIO, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+declare BUILD_VERSION KUBECTL_DIRECTPV
+
+function init() {
+    if [ "$#" -ne 1 ]; then
+        cat <<EOF
+USAGE:
+  helm.sh <VERSION>
+
+EXAMPLE:
+  $ helm.sh 4.0.7
+EOF
+        exit 255
+    fi
+
+    # assign after trimming 'v'
+    BUILD_VERSION="${1/v/}"
+
+    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)"
+    if which "kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)" >/dev/null 2>&1; then
+        KUBECTL_DIRECTPV="kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)"
+    elif [ ! -f "${KUBECTL_DIRECTPV}" ]; then
+        echo "Downloading required kubectl-directpv"
+        curl --silent --location --insecure --fail --output "${KUBECTL_DIRECTPV}" "https://github.com/minio/directpv/releases/download/v${BUILD_VERSION}/${KUBECTL_DIRECTPV:2}"
+        chmod a+x "${KUBECTL_DIRECTPV}"
+    fi
+}
+
+function main() {
+    mkdir -p "./helm/templates/"
+    "${KUBECTL_DIRECTPV}" install -o yaml | sed -e 's/^---/~~~/g' | awk '{f="file" NR; print $0 > f}' RS='~~~'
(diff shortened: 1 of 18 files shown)


directpv-chart/.helmignore
balamurugana, line 1: 1. We could have `helm` directory instead of `directpv-chart`
2. We could have below script as `helm.sh` to generate YAMLs
```sh
#!/bin/bash
# This file is part of MinIO DirectPV
# Copyright (c) 2023 MinIO, Inc.
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU Affero General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU Affero General Public License for more details.
#
# You should have received a copy of the GNU Affero General Public License
# along with this program.  If not, see <http://www.gnu.org/licenses/>.

set -o errexit
set -o nounset
set -o pipefail

declare BUILD_VERSION KUBECTL_DIRECTPV

function init() {
    if [ "$#" -ne 1 ]; then
        cat <<EOF
USAGE:
  helm.sh <VERSION>

EXAMPLE:
  $ helm.sh 4.0.6
EOF
        exit 255
    fi

    # assign after trimming 'v'
    BUILD_VERSION="${1/v/}"

    KUBECTL_DIRECTPV="./kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)"
    if which "kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)" >/dev/null 2>&1; then
        KUBECTL_DIRECTPV="kubectl-directpv_${BUILD_VERSION}_$(go env GOOS)_$(go env GOARCH)"
    elif [ ! -f "${KUBECTL_DIRECTPV}" ]; then
        echo "Downloading required kubectl-directpv"
        curl --silent --location --insecure --fail --output "${KUBECTL_DIRECTPV}" "https://github.com/minio/directpv/releases/download/v${BUILD_VERSION}/${KUBECTL_DIRECTPV:2}"
        chmod a+x "${KUBECTL_DIRECTPV}"
    fi
}

function main() {
    mkdir -p "./helm/templates/"
    "${KUBECTL_DIRECTPV}" install -o yaml | sed -e 's/^---/~~~/g' | awk '{f="file" NR; print $0 > f}' RS='~~~'
    for file in file*; do
        name=$(awk '/^kind:/ { print $NF }' "${file}")
        if [ "${name}" == "CustomResourceDefinition" ]; then
            name=$(awk '/^  name:/ { print $NF }' "${file}")
        fi
        if [ -n "${name}" ]; then
            mv "${file}" "./helm/templates/${name}.yaml"
        fi
    done
}

init "$@"
main "$@"
```

helm.sh
```diff
@@ -0,0 +1,64 @@
...
+declare BUILD_VERSION KUBECTL_DIRECTPV
+
+function init() {
+    if [ "$#" -ne 1 ]; then
+        cat <<EOF
+USAGE:
+  helm.sh <VERSION>
+
+EXAMPLE:
+  $ helm.sh 4.0.6
```
Praveenrajmani, line 31:
```suggestion
  $ helm.sh 4.0.7
```

as v4.0.7 is the latest version as of now

helm/templates/CSIDriver.yaml
```diff
@@ -0,0 +1,19 @@
+
```
Praveenrajmani, line 1: can we remove the top and bottom empty lines in these manifests? just only empty bottom line would be enough

Praveenrajmani (dismissed):
Can we generate the manifests for the latest version (v4.0.7)?

LGTM otherwise

balamurugana (dismissed):
We would need to have configurable values for
* Node selector
* Tolerations
* Apparmor profile
* Seccomp profile
* Image pull secrets
* directpv image
* csi-node-registrar image
* csi-provisioner image
* livenessprobe image
* csi-resizer image
* deployment selector value (for upgrade)
* daemonset selector value (for upgrade)

This can be addressed in separate PRs

helm/templates/CSIDriver.yaml
balamurugana: These files are auto generated.  IMO this is OK.

balamurugana (approved):

Praveenrajmani (approved):

pschichtel:
@cniackz where will the chart be published? minio's main repo?

balamurugana:
@pschichtel There is no plan to support this helm chart. We are primarily working on OpenShift certification for DirectPV and this chart is specifically for that.

pschichtel:
@balamurugana 

from a first glance there doesn't seem to be a reason this chart wouldn't also work on standard kubernetes. from a deployment automation standpoint a helm chart would be great to have.

is this more of a "it's not a priority for now" thing? and would PRs with chart changes be accepted?

balamurugana:
@pschichtel It works but it is not complete. I mentioned few in https://github.com/minio/directpv/pull/821#pullrequestreview-1573247519. We welcome your PRs. Please feel free to send them.

PR #827, opened 2023-08-21
Indicate suspended volumes in 'kubectl directpv list volumes' command

Praveenrajmani:


cmd/kubectl-directpv/list_volumes.go
```diff
@@ -250,6 +250,10 @@ func listVolumesMain(ctx context.Context) {
 			status = "Bounded"
 		}
 
+		if volume.IsSuspended() {
+			status += " (suspended)"
```
balamurugana, line 254: We can have comma separated values here.

PR #829, opened 2023-08-25
Fix minikube setup in Github CI

Praveenrajmani:
fix: `cannot access '/etc/cni/net.d': No such file or directory` while setting minikube for functional tests


.github/workflows/functests.yml
```diff
@@ -57,6 +57,11 @@ jobs:
           docker build --quiet -t quay.io/minio/directpv:${VERSION} .
           docker build --quiet -t example.org/test/sleep:v0.0.1 -f functests/Dockerfile.sleep functests
 
+      - name: Minikube pre-check
+        run: |
+          # To Fix: chmod: cannot access '/etc/cni/net.d': No such file or directory
+          sudo mkdir -p /etc/cni/net.d
```
balamurugana, line 63: You could move it to `Set environment`

balamurugana (approved):

PR #848, opened 2023-09-18
Use SHA digests instead of tags in operator bundle

cniackz:
### Objective:

To use SHA digests instead of tags

### Reasoning:

* [verify-pinned-digest](https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/troubleshooting.md#verify-pinned-digest)

* It is a requirement from RedHat to use SHA Digests.

### DIFF
--- a/operator/bundle/manifests/minio-directpv-operator-rhmp.clusterserviceversion.yaml
+++ b/operator/bundle/manifests/minio-directpv-operator-rhmp.clusterserviceversion.yaml
@@ -222,7 +222,7 @@ spec:
                 - --upstream=http://127.0.0.1:8080/
                 - --logtostderr=true
                 - --v=0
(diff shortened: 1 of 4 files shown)


balamurugana (dismissed):

balamurugana (approved):

Praveenrajmani (approved):
Q: Do we have all the `quay.io/cniackz4/` images pushed to upstream - `quay.io/minio`? @cniackz

cniackz:
Hello @Praveenrajmani 

We are not using `cniackz4` for any image. We are only using `minio` for the images.
But only Admin can generate and push those images.
I don't think we need to change anything else other than push new images on every change.

Regards,
-Cesar

PR #851, opened 2023-09-20
fix legacy DirectCSI upgrade documentation

balamurugana:


docs/upgrade.md
```diff
@@ -21,14 +21,18 @@ Follow the below steps for an in-place upgrade
 $ curl -sfL https://github.com/minio/directpv/raw/master/docs/tools/install.sh | sh - apply
 ```
 
-### Upgrade legacy DirectCSI CSI driver v3.x.x and v2.0.9
+### Upgrade legacy DirectCSI CSI driver
 Follow the below steps to upgrade to the latest DirectPV CSI driver from a legacy DirectCSI installation.
-1. Uninstall legacy DirectCSI CSI driver.
+1. Uninstall DirectCSI driver if you run v3.1.0 or newer version. For other versions, skip this step.
```
Praveenrajmani, line 26: this step can be skipped only when conversion webhook is needed. ie, when there are mixed storage versions and we require conversion webhook to query those mixed versions.

so, can we also add a troubleshooting doc for this. so that the users will know what to do when they see errors like

```
Error; unable to get legacy drives; conversion webhook for direct.csi.min.io/v1beta3, Kind=DirectCSIDrive failed: Post "https://direct-csi-min-io.direct-csi-min-io.svc:30443/convertdrive?timeout=30s":
```

balamurugana: Conversion webhook and mixed directpv CRD version are too much internals to say here. As we say version level like >= v3.1.0 is clear to the users.

However I will add FAQ for this error

Praveenrajmani (approved):

PR #863, opened 2023-09-30
error out for invalid label in create-storage-class.sh

balamurugana:


Praveenrajmani:
let me do some testing before merging

PR #873, opened 2023-10-18
fix drive name of volume at startup if needed

Praveenrajmani:
When a node-server starts, it syncs the drive states like drive name, capacity, make etc. This is because, the drive order wouldv'e changed during node restarts.

In such scenarios, We need to also make sure to fix the drive name in corresponding volumes. Without this fix, the `kubectl directpv list volumes` command will show wrong drive names (if the node has restarted)

### DIFF
--- a/pkg/volume/event.go
+++ b/pkg/volume/event.go
@@ -30,6 +30,7 @@ import (
 	"github.com/minio/directpv/pkg/sys"
 	"github.com/minio/directpv/pkg/types"
 	"github.com/minio/directpv/pkg/xfs"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/cache"
@@ -77,13 +78,35 @@ func (handler *volumeEventHandler) ObjectType() runtime.Object {
 	return &types.Volume{}
 }
 
-func (handler *volumeEventHandler) Handle(ctx context.Context, _ controller.EventType, object runtime.Object) error {
+func (handler *volumeEventHandler) Handle(ctx context.Context, eventType controller.EventType, object runtime.Object) error {
 	volume := object.(*types.Volume)
 	if !volume.GetDeletionTimestamp().IsZero() {
 		return handler.delete(ctx, volume)
 	}
 
+	if eventType == controller.AddEvent {
+		return sync(ctx, volume)
+	}
+
 	return nil
+}
+
+func sync(ctx context.Context, volume *types.Volume) error {
+	drive, err := client.DriveClient().Get(ctx, string(volume.GetDriveID()), metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		return nil
+	}
+	driveName := drive.GetDriveName()
+	if volume.GetDriveName() != driveName {
+		volume.SetDriveName(driveName)
+		_, err = client.VolumeClient().Update(ctx, volume, metav1.UpdateOptions{
+			TypeMeta: types.NewVolumeTypeMeta(),
+		})
+	}
+	return err
 }
 
 func (handler *volumeEventHandler) delete(ctx context.Context, volume *types.Volume) error {
--- a/pkg/device/sync_test.go
+++ b/pkg/device/sync_test.go
@@ -1,0 +1,117 @@
+// This file is part of MinIO DirectPV
+// Copyright (c) 2023 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package device
+
+import (
+	"testing"
+
+	directpvtypes "github.com/minio/directpv/pkg/apis/directpv.min.io/types"
+	"github.com/minio/directpv/pkg/types"
+)
+
+func newDrive(name string, totalCapacity int64, make, volume string) *types.Drive {
+	drive := types.NewDrive(
+		directpvtypes.DriveID(name+"-id"),
+		types.DriveStatus{
+			TotalCapacity: totalCapacity,
+			Make:          make,
+		},
+		directpvtypes.NodeID("nodeId"),
+		directpvtypes.DriveName(name),
+		directpvtypes.AccessTierDefault,
+	)
+	drive.AddVolumeFinalizer(volume)
+	return drive
+}
+
+func newTestDevice(name string, totalCapacity int64, dmname string) device {
+	return device{
+		TotalCapacity: totalCapacity,
+		Device: Device{
+			Name:   name,
+			DMName: dmname,
+		},
+	}
+}
+
+func TestSyncDrive(t *testing.T) {
+	testCases := []struct {
+		drive                 *types.Drive
+		device                device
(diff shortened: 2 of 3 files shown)


Praveenrajmani:
I need to test this on real machines before merging

pkg/device/sync.go
```diff
@@ -74,10 +74,39 @@ func probeDeviceMap() (map[string][]device, error) {
...
+func syncDriveStates(ctx context.Context, drive *types.Drive, device device) (updated bool, err error) {
+	driveName := directpvtypes.DriveName(device.Name)
+	if drive.GetDriveName() != driveName {
+		// This can happen when the drive order changes on node reboots.
 		updated = true
-		drive.SetDriveName(directpvtypes.DriveName(device.Name))
+		drive.SetDriveName(driveName)
+	}
+	// Update drive name in volumes
+	for _, volumeName := range drive.GetVolumes() {
```
balamurugana, line 85: Keep volume update as separate function after `Sync()`

pkg/device/sync.go
```diff
@@ -127,7 +160,13 @@ func Sync(ctx context.Context, nodeID directpvtypes.NodeID) error {
 				if syncDrive(drive, devices[0]) {
 					updated = true
 				}
-
+				driveStateUpdated, err := syncVolume(ctx, drive)
```
balamurugana, line 163: We need to have a separate function outside of drive sync up something like below

```go
func updateDrive(ctx context.Context, driveName string, deviceMap map[string][]device) error {
	drive, err := client.DriveClient().Get(ctx, driveName, metav1.GetOptions{})
	if err != nil {
		if apierrors.IsNotFound(err) {
			return nil
		}
		return err
	}
	var updated bool
	devices := deviceMap[drive.Status.FSUUID]
	switch len(devices) {
	case 0:
		// no match
		if drive.Status.Status != directpvtypes.DriveStatusLost {
			updated = true
			drive.Status.Status = directpvtypes.DriveStatusLost
		}
	case 1:
		// device found
		if syncDrive(drive, devices[0]) {
			updated = true
		}

		// verify mount
		switch drive.Status.Status {
		case directpvtypes.DriveStatusReady, directpvtypes.DriveStatusLost, directpvtypes.DriveStatusError, directpvtypes.DriveStatusMoving:
			source := utils.AddDevPrefix(string(drive.GetDriveName()))
			target := types.GetDriveMountDir(drive.Status.FSUUID)
			if err = xfs.Mount(source, target); err != nil {
				updated = true
				drive.Status.Status = directpvtypes.DriveStatusError
				drive.SetMountErrorCondition(fmt.Sprintf("unable to mount; %v", err))
				// Events and logs
				client.Eventf(drive, client.EventTypeWarning, client.EventReasonDriveMountError, "unable to mount the drive; %v", err)
				klog.ErrorS(err, "unable to mount the drive", "Source", source, "Target", target)
			} else {
				client.Eventf(
					drive,
					client.EventTypeNormal,
					client.EventReasonDriveMounted,
					"Drive mounted successfully to %s", target,
				)

				if drive.Status.Status == directpvtypes.DriveStatusLost {
					updated = true
					drive.Status.Status = directpvtypes.DriveStatusReady
				}

				latestErrorConditionType := drive.GetLatestErrorConditionType()
				if drive.Status.Status == directpvtypes.DriveStatusError {
					switch latestErrorConditionType {
					case directpvtypes.DriveConditionTypeMountError, directpvtypes.DriveConditionTypeMultipleMatches, directpvtypes.DriveConditionTypeRelabelError:
						updated = true
						drive.Status.Status = directpvtypes.DriveStatusReady
					}
				}

				if drive.Spec.Relabel {
					updated = true

					if err = os.Symlink(".", types.GetVolumeRootDir(drive.Status.FSUUID)); err != nil {
						switch {
						case errors.Is(err, os.ErrExist):
							err = nil
						default:
							if latestErrorConditionType != directpvtypes.DriveConditionTypeIOError {
								drive.Status.Status = directpvtypes.DriveStatusError
								drive.SetRelabelErrorCondition(fmt.Sprintf("unable to relabel; %v", err))
							}

							client.Eventf(
								drive,
								client.EventTypeWarning,
								client.EventReasonDriveRelabelError,
								"unable to relabel; %v", err,
							)

							klog.ErrorS(
								err,
								"unable to create symlink",
								"symlink", types.GetVolumeRootDir(drive.Status.FSUUID),
								"drive", drive.Name,
							)
						}
					}

					if err == nil {
						drive.Spec.Relabel = false
					}
				}
			}
		}
	default:
		// more than one matching devices
		updated = true
		var deviceNames string
		for i := range devices {
			if deviceNames != "" {
				deviceNames += ", "
			}
			deviceNames += devices[i].Name
		}
		drive.Status.Status = directpvtypes.DriveStatusError
		drive.SetMultipleMatchesErrorCondition(fmt.Sprintf("multiple devices found by FSUUID; %v", deviceNames))
		client.Eventf(drive, client.EventTypeWarning, client.EventReasonDriveHasMultipleMatches, "unable to mount the drive due to %v", err)
		klog.ErrorS(err, "multiple devices found by FSUUID", "drive", drive.GetDriveName(), "FSUUID", drive.Status.FSUUID, "devices", deviceNames)
	}
	if updated {
		_, err = client.DriveClient().Update(ctx, drive, metav1.UpdateOptions{TypeMeta: types.NewDriveTypeMeta()})
	}
	return err
}

// syncDrives - matches and syncs drives with locally probed devices.
func syncDrives(ctx context.Context, nodeID directpvtypes.NodeID) error {
	deviceMap, err := probeDeviceMap()
	if err != nil {
		return err
	}
	drives, err := drive.NewLister().NodeSelector([]directpvtypes.LabelValue{directpvtypes.ToLabelValue(string(nodeID))}).Get(ctx)
	if err != nil {
		return err
	}
	for i := range drives {
		if err = retry.RetryOnConflict(retry.DefaultRetry, func() error { return updateDrive(ctx, drives[i].Name, deviceMap) }); err != nil {
			return err
		}
	}

	return nil
}

// syncVolumes - matches and syncs drives with locally probed devices.
func syncVolumes(ctx context.Context, nodeID directpvtypes.NodeID) error {
	// FIXME
	return nil
}

// Sync - syncs drives and volumes
func Sync(ctx context.Context, nodeID directpvtypes.NodeID) error {
	derr := syncDrives(ctx, nodeID)
	verr := syncVolumes(ctx, nodeID)
	if derr != nil && verr != nil {
		return fmt.Errorf("%w; %w", derr, verr)
	}
	if derr != nil {
		return derr
	}
	return verr
}
```

Also move `sync.go` outside of `pkg/device` may be `pkg/sync`.

Praveenrajmani: moved this to volume listener Add()

pkg/device/sync.go
```diff
@@ -76,6 +76,7 @@ func probeDeviceMap() (map[string][]device, error) {
 
 func syncDrive(drive *types.Drive, device device) (updated bool) {
 	if string(drive.GetDriveName()) != device.Name {
+		// This can happen when the drive order changes on node reboots.
```
balamurugana, line 79: It is possible to change drive name at runtime as well. You could remove this comment i.e. revert changes in this file.

pkg/volume/event.go
```diff
@@ -77,15 +78,37 @@ func (handler *volumeEventHandler) ObjectType() runtime.Object {
...
 }
 
-func (handler *volumeEventHandler) Handle(ctx context.Context, _ controller.EventType, object runtime.Object) error {
+func (handler *volumeEventHandler) Handle(ctx context.Context, eventType controller.EventType, object runtime.Object) error {
 	volume := object.(*types.Volume)
 	if !volume.GetDeletionTimestamp().IsZero() {
 		return handler.delete(ctx, volume)
 	}
 
+	if eventType == controller.AddEvent {
```
balamurugana, line 87: For existing volumes, `Add` event occurs only at start of the controller. This is same as added into `Sync()`. As volumes live for very long, we could add drive name update every day once in future.

Praveenrajmani: Do you want to keep this in Update event handler? The reason i added this in Add event handler is because, this can happen only during the node restarts mostly. 

Adding this in Update handler, makes these repeated GET calls to the respective drive objects.

If the drive order can change at runtime (considering it as a rare event), can we suggest them to restart the pods instead of checking everytimea at the update handler?

balamurugana: Yep. We will add it later.

balamurugana (approved):

PR #874, opened 2023-10-23
Check and update errors in drive event handling

balamurugana:


pkg/drive/event.go
```diff
@@ -345,10 +350,62 @@ func (handler *driveEventHandler) handleUpdate(ctx context.Context, drive *types
...
+				klog.ErrorS(uerr, "unable to set I/O error", "drive", drive.GetDriveID())
+			}
+		} else {
+			klog.ErrorS(err, "unable to check volume root directory existent", "drive", drive.GetDriveID())
+		}
+
+		return nil
+	}
+
+	if _, err := handler.getDeviceByFSUUID(drive.Status.FSUUID); err != nil {
```
Praveenrajmani, line 370: we should also revert the state back to "Ready" when the device by FSUUID is found

balamurugana: done

pkg/drive/event.go
```diff
@@ -345,10 +350,92 @@ func (handler *driveEventHandler) handleUpdate(ctx context.Context, drive *types
...
+				"unable to find device by FSUUID; "+
+					"either device is removed or run command "+
+					"`sudo udevadm control --reload-rules && sudo udevadm trigger`"+
+					"on the host to reload",
+				"FSUUID", drive.Status.FSUUID)
+			client.Eventf(
+				drive, client.EventTypeWarning, client.EventReasonDeviceNotFoundError,
+				"unable to find device by FSUUID %v; "+
+					"either device is removed or run command "+
+					"`sudo udevadm control --reload-rules && sudo udevadm trigger`"+
```
Praveenrajmani, line 380: actually, `udevadm trigger` will not fix the symlink problems sometime. we may require a reboot in such cases. do we need to communicate this information as well?

balamurugana: No. It is not a general case. `udevadm trigger` is general guideline whereas node reboot is specific case.

Praveenrajmani (approved):

PR #877, opened 2023-10-30
[Proposal] Support moving data from one drive to another within the same node

Praveenrajmani:
This is a feature proposal to support moving of data from one drive to another. Currently, we support moving the references only.

- By adding another flag `--with-data` in `kubectl directpv move` command, creates the job for moving the volumes including data from one drive to another within the same node.
- New jobs will be created which will create pods to move the data from one drive to another.
- Until the job completes, the volumes will be in "moving" state and will not be binded. ie, NodeStageVolume calls will fail/wait. Upon successful completion, the volumes will be binded and the pod will be up.

### DIFF
--- a/cmd/directpv/copy.go
+++ b/cmd/directpv/copy.go
@@ -1,0 +1,237 @@
+// This file is part of MinIO DirectPV
+// Copyright (c) 2023 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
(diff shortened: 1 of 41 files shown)


cmd/directpv/copy.go
```diff
@@ -0,0 +1,234 @@
...
+
+func printProgress(ctx context.Context, srcDrive, destDrive *types.Drive, volume *types.Volume) {
+	sourceQ, err := xfs.GetQuota(ctx, "/dev/"+string(srcDrive.GetDriveName()), volume.Name)
+	if err != nil {
+		klog.ErrorS(err, "unable to get quota of the source drive", "source drive", srcDrive.GetDriveName(), "volume", volume.Name)
+		return
+	}
+	destQ, err := xfs.GetQuota(ctx, "/dev/"+string(destDrive.GetDriveName()), volume.Name)
+	if err != nil {
+		klog.ErrorS(err, "unable to get quota of the destination drive", "destination drive", destDrive.GetDriveName(), "volume", volume.Name)
```
Praveenrajmani, line 171: we need to exit the loop on such quota errors. will fix.

buccella:
Any chance this gets merged before the next release?

Praveenrajmani:
closing this stale PR; will create another PR on demand

PR #885, opened 2023-12-12
change the "fstype" sc parameter to "csi.storage.k8s.io/fstype"

Praveenrajmani:
"fstype" storage class parameter is deprecated

### DIFF
--- a/docs/tools/create-storage-class.sh
+++ b/docs/tools/create-storage-class.sh
@@ -82,7 +82,7 @@ metadata:
     directpv.min.io/version: v1beta1
   name: ${NAME}
 parameters:
-  fstype: xfs
+  csi.storage.k8s.io/fstype: xfs
 $(printf '  %s\n' "${DRIVE_LABELS[@]}")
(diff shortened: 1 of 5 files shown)


docs/tools/create-storage-class.sh
```diff
@@ -82,7 +82,7 @@ metadata:
     directpv.min.io/version: v1beta1
   name: ${NAME}
 parameters:
-  fstype: xfs
+  csi.storage.k8s.io/fstype: xfs
```
harshavardhana, line 85: Will this be a harmless change?

Praveenrajmani: yes, this is harmless. 

I have seen logs like `"fstype" is deprecated and will be removed in a future release, please use "csi.storage.k8s.io/fstype" instead` in older k8s versions

harshavardhana: okay wonderful.

harshavardhana (approved):

balamurugana (approved):

PR #906, opened 2024-04-22
fix: handle Add event in node event listener

Praveenrajmani:
bonus: also fix the nil value while creating the node during refresh during `kubectl directpv discover` command

**Steps to test** 

- Delete a directpvnode using `kubectl delete directpvnode <node-name>`
- Try `kubectl directpv discover`

### DIFF
--- a/.github/workflows/functests.yml
+++ b/.github/workflows/functests.yml
@@ -20,7 +20,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        kube-version: ['v1.18.20', 'v1.19.16', 'v1.20.15', 'v1.21.14', 'v1.22.17', 'v1.23.17', 'v1.24.17', 'v1.25.16', 'v1.26.14', 'v1.27.11', 'v1.28.7', 'v1.29.2']
+        kube-version: ['v1.20.15', 'v1.21.14', 'v1.22.17', 'v1.23.17', 'v1.24.17', 'v1.25.16', 'v1.26.15', 'v1.27.13', 'v1.28.9', 'v1.29.4', 'v1.30.0']
         os: [ubuntu-20.04, ubuntu-22.04]
         exclude:
           - os: ubuntu-22.04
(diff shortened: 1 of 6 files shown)


.github/workflows/functests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     timeout-minutes: 60
     strategy:
       matrix:
-        kube-version: ['v1.18.20', 'v1.19.16', 'v1.20.15', 'v1.21.14', 'v1.22.17', 'v1.23.17', 'v1.24.17', 'v1.25.16', 'v1.26.14', 'v1.27.11', 'v1.28.7', 'v1.29.2']
+        kube-version: ['v1.20.15', 'v1.21.14', 'v1.22.17', 'v1.23.17', 'v1.24.17', 'v1.25.16', 'v1.26.14', 'v1.27.11', 'v1.28.7', 'v1.29.2', 'v1.30.0']
```
balamurugana, line 23:
```suggestion
        kube-version: ['v1.20.15', 'v1.21.14', 'v1.22.17', 'v1.23.17', 'v1.24.17', 'v1.25.16', 'v1.26.15', 'v1.27.13', 'v1.28.9', 'v1.29.4', 'v1.30.0']
```

balamurugana, line 23: Also fix https://github.com/minio/directpv/blob/master/.github/workflows/functests.yml#L85

pkg/installer/installer.go
```diff
@@ -88,7 +88,7 @@ func Install(ctx context.Context, args *Args) (err error) {
 	case args.DryRun:
 		if args.KubeVersion == nil {
 			// default higher version
-			if args.KubeVersion, err = version.ParseSemantic("1.27.0"); err != nil {
+			if args.KubeVersion, err = version.ParseSemantic("1.30.0"); err != nil {
```
balamurugana, line 91: 1. `v1.30.0` is just a week old and we don't want to force to use it. Instead you could stick to `v1.29.0`
2. You could fix https://github.com/minio/directpv/blob/master/cmd/kubectl-directpv/install.go#L53

pkg/installer/installer.go
```diff
@@ -111,8 +111,8 @@ func Install(ctx context.Context, args *Args) (err error) {
 	}
 
 	if args.KubeVersion.Major() != 1 ||
-		args.KubeVersion.Minor() < 18 ||
-		args.KubeVersion.Minor() > 27 {
+		args.KubeVersion.Minor() < 20 ||
+		args.KubeVersion.Minor() > 30 {
```
balamurugana, line 115: As current code works for `v1.18`, you could leave minor version as is. We will remove the support when PSP is removed completely.

pjuarezd (dismissed):
Other than the Bala comments, looks good to me

balamurugana (approved):

pjuarezd (approved):

PR #909, opened 2024-05-13
use genericclioptions package to handle kubeconfig

### LINKED ISSUE #884: KUBECONFIG with several config files `:` separated is not supported
```
kubectl directpv install
F1201 14:22:00.391814    5754 init.go:43] unable to get kubernetes configuration; unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
```
```
% export KUBECONFIG=/Users/gms/.kube/mbbf-minio.yaml
gms@fluffy ~ % kubectl directpv install                          

 ███████████████████████████████████████████████████████████████████████████ 100%

┌──────────────────────────────────────┬──────────────────────────┐
│ NAME                                 │ KIND                     │
├──────────────────────────────────────┼──────────────────────────┤
│ directpv                             │ Namespace                │
│ directpv-min-io                      │ ServiceAccount           │
│ directpv-min-io                      │ ClusterRole              │
│ directpv-min-io                      │ ClusterRoleBinding       │
│ directpv-min-io                      │ Role                     │
│ directpv-min-io                      │ RoleBinding              │
│ directpv-min-io                      │ PodSecurityPolicy        │
│ psp-directpv-min-io                  │ ClusterRoleBinding       │
│ directpvdrives.directpv.min.io       │ CustomResourceDefinition │
│ directpvvolumes.directpv.min.io      │ CustomResourceDefinition │
│ directpvnodes.directpv.min.io        │ CustomResourceDefinition │
│ directpvinitrequests.directpv.min.io │ CustomResourceDefinition │
│ directpv-min-io                      │ CSIDriver                │
│ directpv-min-io                      │ StorageClass             │
│ node-server                          │ Daemonset                │
│ controller                           │ Deployment               │
└──────────────────────────────────────┴──────────────────────────┘

DirectPV installed successfully
```

```
kubectl directpv --version
directpv version v4.0.6
```

bewing:
Use the same flagset that kubectl uses to select context and namespace, and generate the clientset from that.

Fixes #884


bewing:
I believe all codepaths to client init are adjusted now.

cmd/directpv/main.go
```diff
@@ -94,7 +95,9 @@ func init() {
 	flag.Set("logtostderr", "true")
 	flag.Set("alsologtostderr", "true")
 
-	mainCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", kubeconfig, "Path to the kubeconfig file to use for Kubernetes requests.")
+	genericOptions = genericclioptions.NewConfigFlags(true)
+	genericOptions.AddFlags(mainCmd.PersistentFlags())
```
balamurugana, line 99: This is not going to work. We use flags which colliding with `genericcliflags`. In fact we don't need support flags like `kubectl`.

balamurugana (changes requested):

cmd/directpv/main.go
bewing: > We use flags which colliding with genericcliflags

Then use the flags from the flagset where they collide?  I'm happy to work with you to identify where this is and see if it's trivial to access it via the flagset.  I'm not sure if viper is able to access it as I haven't tested

> In fact we don't need support flags like kubectl.

As referenced by #884, your tool does not work for those that are working with multiple kubeconfig files very easily.  The flagset and its associated loaders provide a much more robust cluster and namespace selection, rather than having to juggle multiple kubeconfig directories.

bewing: > We use flags which colliding with genericcliflags

Apologies -- looking at the test output, it's not that you're using the same flags, there's reused shortnames.  That is more of an issue.

balamurugana:
User should solve this issue externally. We do not support directly in DirectPV.

PR #918, opened 2024-07-23
Remove unsupported PodSecurityPolicy

Praveenrajmani:
PSP is deprecated from v1.21 and removed from k8s v1.24. The dependent pkgs have already removed the PSP support. It is time to remove PSP and stop supporting it. 

To be added in release notes 

### No support for PSP from v4.1.x

As [PodSecurityPolicy](https://kubernetes.io/docs/concepts/security/pod-security-policy/) was deprecated in kubernetes v1.21 and removed from kubernetes in v1.25. Its support will be removed from DirectPV v4.1.x.

If you wish to retain the support during the upgrade from versions < v4.1.x to v4.1.x, please follow the steps below

#### STEP 1: Take a backup before upgrade

```sh
> kubectl get psp directpv-min-io -o yaml > psp.yaml
> kubectl get clusterrolebinding psp-directpv-min-io -o yaml > psp-crb.yaml
```

#### STEP 2: Apply the backup after upgrade

```sh
> kubectl apply -f psp.yaml
> kubectl apply -f psp-crb.yaml
```

NOTE: The release will be a minor version bump - `v4.1.0`


docs/upgrade.md
```diff
@@ -44,3 +44,25 @@ $ kubectl krew upgrade directpv
 
 ### Upgrade of release binary
 Refer to the [binary installation documentation](./installation.md#installation-of-release-binary).
+
+## Deprecations
```
balamurugana, line 48: 1. I think we don't need to keep this in the documentation of `v4.1`. This should go into release notes.
2. The release note should indicate `PodSecurityPolicy` removal (not deprecation) as per Kubernetes.
3. We do not need to document how to backup/apply `PodSecurityPolicy` anywhere as it is complete removal from DirectPV.
4. We would need to fix `docs/tools/install.sh` script as well by aborting it if `PodSecurityPolicy` used in case of upgrade.

Praveenrajmani: `docs/tools/install.sh` will use `kubectl directpv install --declarative` which will  not generate PSP manifests @balamurugana

balamurugana: In case of upgrade, `install.sh` will upgrade everything except previously created PSP. As we are removing PSP support completely, we would error out for the case i.e. we fail to upgrade with a message saying use the latest version of `v4.0`

docs/tools/install.sh
```diff
@@ -112,6 +112,17 @@ function check_directcsi_consistency() {
     fi
 }
 
+function is_greater_than_or_equal_to_4.1.0() {
```
balamurugana, line 115: You could add the support like below
```diff
diff --git a/docs/tools/install.sh b/docs/tools/install.sh
index 8d38bdb1..2ebcdb4f 100755
--- a/docs/tools/install.sh
+++ b/docs/tools/install.sh
@@ -28,6 +28,10 @@ function run_plugin_cmd() {
     "${plugin_cmd[@]}" "$@"
 }
 
+function is_psp_found() {
+    kubectl get --ignore-not-found=true psp directpv-min-io --no-headers -o NAME | grep -q .
+}
+
 function is_legacy_drive_found() {
     if ! kubectl get --ignore-not-found=true crd directcsidrives.direct.csi.min.io --no-headers -o NAME | grep -q .; then
         return 1
@@ -180,6 +184,13 @@ function wait_for_crd() {
 }
 
 function main() {
+    if is_psp_found; then
+        echo "[WARNING] Older PodSecurityPolicy found !!!"
+        echo "[WARNING] DirectPV removed PodSecurityPolicy support as per Kubernetes supportability"
+        echo "[WARNING] This installation does not change anything related to existing PodSecurityPolicy"
+        echo "[WARNING] You have to remove it manually and use Pod Security Admission as per your requirement"
+    fi
+
     echo "* Probe legacy DirectCSI"
     legacy=0
     migrate=0
```

Praveenrajmani: Don't we need the directpv version check? sometimes users will just take the script from master and use.

balamurugana: `master` always has the latest code and users should not do that for older releases.

balamurugana (approved):

prakashsvmx (approved):

PR #931, opened 2024-08-22
Fix size probe for invalid devices

Praveenrajmani:
for any invalid (or) detached drives, size probe will fail with the following error

```
strconv.ParseUint: parsing "": invalid syntax
```

which prevents the directpv containers to start

bonus: improve probe error messages

### DIFF
--- a/pkg/device/probe_linux.go
+++ b/pkg/device/probe_linux.go
@@ -20,6 +20,7 @@ package device
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/minio/directpv/pkg/sys"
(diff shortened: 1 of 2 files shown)


pkg/device/probe_linux.go
```diff
@@ -52,31 +53,31 @@ func newDevice(
 	}
 
 	if device.Size, err = getSize(name); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get size for device %v; %v", name, err)
```
balamurugana, line 56: 1. Use `%w` for `err` 
2. you could format it like `unable to get size; device=%v; err=%w`

pkg/device/probe_linux.go
```diff
@@ -52,31 +53,31 @@ func newDevice(
...
 	if device.Size, err = getSize(name); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get size; device=%v; err=%w", name, err)
 	}
 
 	device.Hidden = getHidden(name)
 
 	if device.Removable, err = getRemovable(name); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get removable info; device=%v; err=%w", name, err)
```
balamurugana, line 62: You could use `flag` instead of `info`

pkg/device/probe_linux.go
```diff
@@ -52,31 +53,31 @@ func newDevice(
...
 	partitions, err := getPartitions(name)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get partition info; device=%v; err=%w", name, err)
 	}
 	device.Partitioned = len(partitions) != 0
 
 	if device.DMName, err = getDMName(name); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get dmname info; device=%v; err=%w", name, err)
```
balamurugana, line 80: may be `DM name` instead of `dmname info`

pkg/device/probe_linux.go
```diff
@@ -85,22 +86,22 @@ func newDevice(
...
 	cdroms, err := getCDROMs()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get CDROM information; %w", err)
 	}
 
 	swaps, err := getSwaps()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to get SWAP information; %w", err)
```
balamurugana, line 104: SWAP/swap

pkg/device/probe_linux.go
```diff
@@ -117,17 +118,17 @@ func probe() (devices []Device, err error) {
...
```
balamurugana, line 131: SWAP/swap

balamurugana (approved):

PR #957, opened 2024-10-06
Configure replicas

varac:
This MR makes the controller replicas configurable. 

It also: 

* removes unused values from `values.yaml`
* Removes the `Namespace.yaml` to match Helm best practices (see https://github.com/helm/helm/issues/5465#issuecomment-473942223 for the explanation)


balamurugana (changes requested):
Use https://github.com/minio/directpv/tree/master/resources instead of unmaintained helm and we don't have plans to keep it

PR #980, opened 2024-11-28
De not set XFS retry_timeout_seconds as max_retries is zero

sathieu:
if max_retries is 0, then there's no point in setting retry_timeout_seconds


sathieu:
See original discussion in https://github.com/minio/docs/pull/1369.

Praveenrajmani (approved):

